PR #18, opened 2024-07-31
Add thinning

aloctavodia:
Adds thinning based on ESS as discussed [Säilynoja 2022](https://arxiv.org/abs/2103.10522) and this vignette https://mc-stan.org/posterior/reference/thin_draws.html

I tried with a couple of examples and it works, but not sure I am adding all that is needed. Or if everything is in the right place

----
📚 Documentation preview 📚: https://arviz-stats--18.org.readthedocs.build/en/18/

### DIFF
--- a/src/arviz_stats/base/dataarray.py
+++ b/src/arviz_stats/base/dataarray.py
@@ -202,13 +202,21 @@ class BaseDataArray:
 
         if factor == "auto":
             n_samples = da.sizes["chain"] * da.sizes["draw"]
-            ess_ave = np.minimum(self.ess(da, method="bulk"), self.ess(da, method="tail")).mean()
+            ess_ave = np.minimum(
+                self.ess(da, method="bulk", dims=["chain", "draw"]),
+                self.ess(da, method="tail", dims=["chain", "draw"]),
+            ).mean()
             factor = int(np.ceil(n_samples / ess_ave))
             dims = "draw"
 
         elif isinstance(factor, (float | int)):
             if dims is None:
                 dims = rcParams["data.sample_dims"]
+            if not isinstance(dims, str):
+                if len(dims) >= 2:
+                    raise ValueError("dims must be of length 1")
+                if len(dims) == 1:
+                    dims = dims[0]
 
             factor = int(factor)
             if factor == 1:
--- a/src/arviz_stats/base/dataarray.py
+++ b/src/arviz_stats/base/dataarray.py
@@ -201,7 +201,9 @@ class BaseDataArray:
         n_samples = da.sizes["chain"] * da.sizes["draw"]
 
         if factor == "auto":
-            ess_ave = np.minimum(self.ess(da, method="bulk"), self.ess(da, method="tail")).mean()
+            ess_ave = np.minimum(
+                self.ess(da, method="bulk", dims=dims), self.ess(da, method="tail", dims=dims)
+            ).mean()
             factor = int(np.ceil(n_samples / ess_ave))
 
         elif isinstance(factor, (float | int)):
@@ -211,7 +213,7 @@ class BaseDataArray:
             if factor < 1:
                 raise ValueError("factor must be greater than 1")
 
-        return da.sel(draw=slice(None, None, factor))
+        return da.sel({dims[-1]: slice(None, None, factor)})
 
 
 dataarray_stats = BaseDataArray(array_class=array_stats)
--- a/src/arviz_stats/base/dataarray.py
+++ b/src/arviz_stats/base/dataarray.py
@@ -212,11 +212,6 @@ class BaseDataArray:
                 raise ValueError("factor must be greater than 1")
 
         return da.sel(draw=slice(None, None, factor))
-        # return apply_ufunc(
-        #     self.array_class.thin,
-        #     da,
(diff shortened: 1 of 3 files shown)


src/arviz_stats/base/dataarray.py
```diff
@@ -193,5 +193,30 @@ def kde(self, da, dims=None, circular=False, grid_len=512, **kwargs):
         out = concat((grid, pdf), dim=plot_axis)
         return out.assign_coords({"bw" if da.name is None else f"bw_{da.name}": bw})
 
+    def thin(self, da, factor="auto", dims=None):
```
aloctavodia, line 196: It is ok to have this logic here, or it should be inside src/arviz_stats/base/array.py (see commented code). Maybe ok because we are just slicing the DataArray so no need to have ufuncs and all that???

OriolAbril: I think here it is preferable to implement the function here. If eventually there are libraries actually using the array only backend we can implement an independent version there and not use it here. The reason why is if we do this via the array backend, the dimension being modified (generally `draw` here) will have to be moved to the last position. Dimension order should be taken for granted in xarray, but it is also nice to not change the order unless necessary, and here it isn't

src/arviz_stats/base/dataarray.py
```diff
@@ -193,5 +193,30 @@ def kde(self, da, dims=None, circular=False, grid_len=512, **kwargs):
...
+        """Compute eti on DataArray input."""
+        if dims is None:
+            dims = rcParams["data.sample_dims"]
+
+        n_samples = da.sizes["chain"] * da.sizes["draw"]
+
+        if factor == "auto":
+            ess_ave = np.minimum(self.ess(da, method="bulk"), self.ess(da, method="tail")).mean()
+            factor = int(np.ceil(n_samples / ess_ave))
```
OriolAbril, line 205:
```suggestion
        if factor == "auto":
            n_samples = np.product(da.sizes[dim] for dim in sample_dims)
            ess_ave = np.minimum(self.ess(da, method="bulk", dims=sample_dims), self.ess(da, method="tail", dims=sample_dims)).mean()
            factor = int(np.ceil(n_samples / ess_ave))
```

I would use the same logic here as in ess/mcse/rhat which basically boils down to `sample_dims` must be length 2, first one is "chain" the 2nd is "draw". And now that I write it we'll have to fix it, only rhat should do this, ess and mcse should work on 1d arrays/single dimension too.

A potential fix could be to do something closer to the array backend, where there is a `chain_axis` and a `draw_axis` but using `_dim` instead.

By default both are none, then:
  - sample_dims is a single dimension: chain_dim continues to be none, draw_dim=sample_dims
  - sample_dims is length 2: chain_dim=sample_dims[0], draw_dim=sample_dims[1]
  - sample_dims is length>2: error

Thinning then happens along `draw_dim` (modifying comment below).

src/arviz_stats/base/dataarray.py
```diff
@@ -193,5 +193,30 @@ def kde(self, da, dims=None, circular=False, grid_len=512, **kwargs):
...
+
+        elif isinstance(factor, (float | int)):
+            factor = int(factor)
+            if factor == 1:
+                return da
+            if factor < 1:
+                raise ValueError("factor must be greater than 1")
+
+        return da.sel(draw=slice(None, None, factor))
```
OriolAbril, line 214:
```suggestion
        return da.sel({sample_dims[-1]: slice(None, None, factor)})
```

maybe something like this? I think this would work even after fixing the logic in ess.

OriolAbril:
This is a very nice feature!

src/arviz_stats/base/dataarray.py
```diff
@@ -193,5 +195,28 @@ def kde(self, da, dims=None, circular=False, grid_len=512, **kwargs):
...
         return out.assign_coords({"bw" if da.name is None else f"bw_{da.name}": bw})
 
+    def thin(self, da, factor="auto", dims=None):
+        """Perform thinning on DataArray input."""
+        if factor == "auto" and dims is not None:
+            warnings.warn("dims are ignored if factor is auto")
+
+        if factor == "auto":
+            n_samples = da.sizes["chain"] * da.sizes["draw"]
+            ess_ave = np.minimum(self.ess(da, method="bulk"), self.ess(da, method="tail")).mean()
```
OriolAbril, line 205: I think we should call ess with dims=["chain", "draw"] othereise the behaviour would be unexpected oe broken if the sample_dims rcparams is modified

src/arviz_stats/base/dataarray.py
```diff
@@ -193,5 +195,28 @@ def kde(self, da, dims=None, circular=False, grid_len=512, **kwargs):
...
+            factor = int(np.ceil(n_samples / ess_ave))
+            dims = "draw"
+
+        elif isinstance(factor, (float | int)):
+            if dims is None:
+                dims = rcParams["data.sample_dims"]
```
OriolAbril, line 211: We should check if after that dims is not a string, if so check the length >=2 should be an error, =1we should get the element inside the list

OriolAbril (commented):
Left two small comments, should be ready to merge after that

PR #20, opened 2024-08-08
Add khat and ps_min_ss

aloctavodia:
This is just a placeholder, the goal is to eventually have a function to compute the `min_ss` as discussed [here](https://mc-stan.org/posterior/reference/pareto_diags.html). The original implementation is [here](https://github.com/stan-dev/posterior/blob/20b6d3aad2d15e398dbf86787da01843f824df7f/R/pareto_smooth.R). There a lot of other functions implemented  (also discussed in the vignette from the first link), I think we don't want those other functions, at least not now.

I checked that I am getting the same result as in the vignette (to be sure I installed the Posterior package, I got the value of mu and converted it into a DataArray)

```
mu <- extract_variable_matrix)(example_draws(), "mu")
pareto_diags(mu)
#> $khat
#> [1] 0.1979001
#> 
#> $min_ss
#> [1] 17.6493
```

This is an example of how to run the function, first compute the k-hat and then the min_ss

```python
sc = az.load_arviz_data("centered_eight")

mu = sc.posterior["mu"]
k = pareto_khat(mu)
ps_min_ss(k)
```

Notice that a lot of this code overlaps with the code necessary to compute the pareto smoothing part of the PSIS-CV-LOO method (notice the "log_weights" argument in the code in this PR). So eventually we will need to use some of this code for LOO too.

----
📚 Documentation preview 📚: https://arviz-stats--20.org.readthedocs.build/en/20/


src/arviz_stats/base/pareto.py
OriolAbril, line 1: All the functions here should be methods in one of the classes. Not sure if core or diagnostics. We should probably define and write down the dependency among classes. For example:

```
                 Core
                /    \
         Density      Diagnostics
                           |
                         Stats
```

so things that are needed everywhere go in core, things needed by only diagnostics and stats go in diagnostics. Or add more classes to keep things better organized. But the idea is for another computational backend to inherit from that, and if there `_gpdfit` is modified to make it faster or test an alternative algorithm, the functions that use `_gpdfit` should not need to be updated. If we put these outside the class and don't use `self.xyz` to achieve that it would be necessary to modify `pareto_khat`, `ps_tail` and `pareto_min_ss` to actually be able to modify `_gpdfit`

src/arviz_stats/base/diagnostics.py
```diff
@@ -327,3 +328,19 @@ def _mcse_quantile(self, ary, prob):
...
         return (th2 - th1) / 2
+
+    def _pareto_min_ss(self, ary):
+        """Compute the minimum effective sample size."""
+        ary = np.asarray(ary)
+        if _not_valid(ary, shape_kwargs={"min_draws": 4, "min_chains": 1}):
+            return np.nan
+        ary_flatten = ary.flatten()
+        n_draws = len(ary_flatten)
+        r_eff = self._ess_tail(ary) / n_draws
```
OriolAbril, line 339:
```suggestion
        r_eff = self._ess_tail(ary, relative=True)
```

src/arviz_stats/base/dataarray.py
```diff
@@ -240,5 +240,16 @@ def thin(self, da, factor="auto", dims=None):
 
         return da.sel({dims: slice(None, None, factor)})
 
+    def pareto_min_ss(self, da, dims=None):
```
OriolAbril, line 243: The array function already supports 1d input (aka only `sample` dimension instead of `chain, draw`). But this one won't. I tried adding support for that in `ess` at https://github.com/arviz-devs/arviz-stats/pull/22/files to unblock `plot_ess`, do you think it is ok to use that convention with `dims` being either length 1 or 2?

If so I could also make a helper/validate function to take care of all that in the other PR and merge it, then it can be used here.

src/arviz_stats/base/array.py
```diff
@@ -132,6 +132,16 @@ def mcse(self, ary, chain_axis=-2, draw_axis=-1, method="mean", prob=None):
...
         return mcse_array(ary, **func_kwargs)
 
+    def pareto_min_ss(self, ary, chain_axis=-2, draw_axis=-1):
+        """Compute minimum effective sample size."""
+        if chain_axis is None:
+            ary = np.expand_dims(ary, axis=0)
+            chain_axis = 0
+        ary, _ = process_ary_axes(ary, [chain_axis, draw_axis])
+        pms_func = getattr(self, "_pareto_min_ss")
+        pms_array = make_ufunc(pms_func, n_output=1, n_input=1, n_dims=2, ravel=False)
```
OriolAbril, line 142:
```suggestion
        pms_array = make_ufunc(self._pareto_min_ss, n_output=1, n_input=1, n_dims=2, ravel=False)
```

method to be "ufuncified" is not dynamic here so it can be accessed directly

src/arviz_stats/base/dataarray.py
aloctavodia: ok

OriolAbril (approved):
Let's merge this first, then rebase the 1d ess PR on top of that and add support for 1d arrays to ess/mcse/pareto_min_ss everywhere in that PR

PR #24, opened 2024-09-19
Add prior sense related functionality

### LINKED ISSUE #23: Add priorsense related functionality
- [ ] psens
- [ ] psense_summary

aloctavodia:
Closes #23

From the user perspective the main addition in this PR is the `psens_summary` function, which is equivalent to [powerscale_sensitivity](https://n-kall.github.io/priorsense) from the priorsense package.

```
azs.psens_summary(sc)

	 prior	likelihood	diagnosis
mu      0.413	0.584	prior-data conflict
sigma	0.297	0.538	prior-data conflict
```

Where `sc` is a DataTree from a PyMC model equivalent to `univariate_normal` model in priorsense docs. Results are close to those, differences may be do to sampling differences, I need to check more carefully. There are also small differences compared to the result from `az.psens ` (now the samples are the same).

----
📚 Documentation preview 📚: https://arviz-stats--24.org.readthedocs.build/en/24/


src/arviz_stats/psens.py
```diff
@@ -0,0 +1,213 @@
...
+        dataset = dataset.sel(coords)
+
+    if group == "log_likelihood":
+        component_draws = get_log_likelihood(dt, var_name=component_var_names, single_var=False)
+    elif group == "log_prior":
+        component_draws = get_log_prior(dt, var_names=component_var_names)
+    else:
+        raise ValueError("Value for `group` argument not recognized")
+
+    component_draws = component_draws.stack(__sample__=("chain", "draw"))
```
OriolAbril, line 89:
```suggestion
    component_draws = component_draws.stack(sample=("chain", "draw"))
```

No more `__sample__` in arviz-stats. `az.extract` already uses `sample`, and we have also extended the "reserved" dimensions across arviz-base and arviz-plots. We just need to be sure to document which dimension names are assumed to have a specific meaning in arviz.

src/arviz_stats/psens.py
```diff
@@ -0,0 +1,213 @@
...
+    and determining the degree to which the posterior changes as described in [1]_.
+    It uses Pareto-smoothed importance sampling to avoid refitting the model.
+
+    References
+    ----------
+    .. [1] Kallioinen et al, *Detecting and diagnosing prior and likelihood sensitivity with
+       power-scaling*, 2022, https://arxiv.org/abs/2107.14054
+    """
+    dataset = extract(dt, var_names=var_names, filter_vars=filter_vars, group="posterior")
+    if coords is None:
```
OriolAbril, line 79:
```suggestion
    if coords is not None:
```

src/arviz_stats/psens.py
```diff
@@ -0,0 +1,213 @@
...
+
+    if group == "log_likelihood":
+        component_draws = get_log_likelihood(dt, var_name=component_var_names, single_var=False)
+    elif group == "log_prior":
+        component_draws = get_log_prior(dt, var_names=component_var_names)
+    else:
+        raise ValueError("Value for `group` argument not recognized")
+
+    component_draws = component_draws.stack(__sample__=("chain", "draw"))
+    if component_coords is None:
```
OriolAbril, line 90:
```suggestion
    if component_coords is not None:
```

src/arviz_stats/psens.py
```diff
@@ -0,0 +1,213 @@
...
+
+        return "-"
+
+    psens_df["diagnosis"] = psens_df.apply(_diagnose, axis=1)
+
+    return psens_df.round(round_to)
+
+
+# get_log_likelihood and get_log_prior functions should be somewhere else
+def get_log_likelihood(idata, var_name=None, single_var=True):
```
OriolAbril, line 176: we have a get log likelihood in https://github.com/arviz-devs/arviz-stats/blob/main/src/arviz_stats/utils.py#L41 already. `utils` is a bad name though

src/arviz_stats/psens.py
```diff
@@ -0,0 +1,213 @@
...
+    elif group == "log_prior":
+        component_draws = get_log_prior(dt, var_names=component_var_names)
+    else:
+        raise ValueError("Value for `group` argument not recognized")
+
+    component_draws = component_draws.stack(__sample__=("chain", "draw"))
+    if component_coords is None:
+        component_draws = component_draws.sel(component_coords)
+    if isinstance(component_draws, xr.DataArray):
+        component_draws = component_draws.to_dataset()
```
OriolAbril, line 93: I'd prefer making sure both `get_log_prior` and `get_log_likelihood` always return a dataset. I'll try to comment below if I see how to easily do this.

src/arviz_stats/psens.py
```diff
@@ -0,0 +1,213 @@
...
+    if len(component_draws.dims):
+        component_draws = component_draws.to_stacked_array(
+            "latent-obs_var", sample_dims=("__sample__",)
+        ).sum("latent-obs_var")
+
+    component_draws = component_draws.unstack()
```
OriolAbril, line 99: not sure what is going on here but that is an extremely weird combination. I'll try to run the code and see if I can make it a bit more xarray-y. I think the combination of code from current arviz (where everything gets squished into 2d arrays) with the use of accessors and dataarray based functions more "native" to arviz-stats is generating some problems.

src/arviz_stats/psens.py
```diff
@@ -0,0 +1,213 @@
...
+
+    # calculate lower and upper alpha values
+    lower_alpha = 1 / (1 + delta)
+    upper_alpha = 1 + delta
+
+    # calculate importance sampling weights for lower and upper alpha power-scaling
+    lower_w = np.exp(component_draws.azstats.power_scale_lw(alpha=lower_alpha)).values.flatten()
+    lower_w = lower_w / np.sum(lower_w)
```
OriolAbril, line 107: I have always been a bit shocked by this. Asssuming `lower_w` comes from the log prior and is squished into a 2d array of shape `(n_variables, n_samples)` do we really want to sum over both dimensions?

src/arviz_stats/psens.py
```diff
@@ -0,0 +1,213 @@
...
+
+    upper_w = np.exp(component_draws.azstats.power_scale_lw(alpha=upper_alpha)).values.flatten()
+    upper_w = upper_w / np.sum(upper_w)
+
+    return dt.azstats.power_scale_sens(
```
OriolAbril, line 112: I think this should be 

```suggestion
    return dataset.azstats.power_scale_sens(
```

because we have used extract and filtered that on `var_names` and `coords`. This would also make the returned object a dataset which I think makes more sense here.

src/arviz_stats/psens.py
```diff
@@ -0,0 +1,213 @@
...
+    psens_df : DataFrame
+        DataFrame containing the prior and likelihood sensitivity values for each variable
+        in the data. And a diagnosis column with the following values:
+        - "prior-data conflict" if both prior and likelihood sensitivity are above threshold
+        - "strong prior / weak likelihood" if the prior sensitivity is above threshold
+        and the likelihood sensitivity is below the threshold
+        - "-" otherwise
+    """
+    pssdp = psens(data, group="log_prior")["posterior"].to_dataset()
+    pssdl = psens(data, group="log_likelihood")["posterior"].to_dataset()
```
OriolAbril, line 142:
```suggestion
    pssdp = psens(data, group="log_prior")
    pssdl = psens(data, group="log_likelihood")
```

with the change above on the outpur of `psens`

src/arviz_stats/psens.py
```diff
@@ -0,0 +1,213 @@
...
+        (np.full((cast(int, n_vars), 2), np.nan)), columns=["prior", "likelihood"]
+    )
+
+    indices = []
+    for i, (var_name, sel, isel, values) in enumerate(
+        xarray_var_iter(joined, skip_dims={"component"})
+    ):
+        psens_df.iloc[i] = values
+        indices.append(labeller.make_label_flat(var_name, sel, isel))
+    psens_df.index = indices
```
OriolAbril, line 160: I opened an issue for a more arvizy version of `to_stacked_array`: https://github.com/arviz-devs/arviz-base/issues/15. I think we also need something like that for dataframe output

src/arviz_stats/psens.py
```diff
@@ -0,0 +1,213 @@
...
+    return log_likelihood
+
+
+def get_log_prior(idata, var_names=None):
+    """Retrieve the log prior dataarray of a given variable."""
+    if not hasattr(idata, "log_prior"):
+        raise TypeError("log prior not found in inference data object")
+    if var_names is None:
+        var_names = list(idata.log_prior.data_vars)
+    return idata.log_prior.to_dataset()[var_names]
```
OriolAbril, line 213:
```suggestion
# somewhere in the imports: from collections.abc import Hashable
    if var_names is None:
        return idata.log_prior
    if isinstance(var_names, Hashable):
        return idata.log_prior[[var_names]]
    return idata.log_prior.to_dataset()[var_names]
```

I think this will guarantee dataset output. The isinstance hashable is a fancier version of checking for string given `dataset[("var", "name")]` is valid given there is a variable named `("var", "name")` (tuples can be variable names too) but it is not valid if there are two variables, one "var" the other "name" in which case `ds[["var", "name"]]` is valid.

We might also want to make a helper function for that as it is quite common. I don't think it would fit in the validation module though, maybe we could have a scaffolding module too? cc @amaloney

amaloney: We certainly can create a scaffolding module.

I am confused as to why are we checking to see if `var_names` is hashable, and more generally, what types do we expect `var_names` to have? I think I've only ever encountered them being strings or tuples. Yes a tuple is hashable, but I think it would be easier to understand if we had a type for `var_names` so that when we checked

```python
if isinstance(var_names, VarNames): ...
```

it would make semantic sense. Then we can define `VarNames` using pydantic, or a dataclass, or some of the newer features in 3.12

OriolAbril: It is either of these 3: None (100% ours to indicate no filtering), `Hashable` -> select a single variable (inherited from xarray, and if you do `dataset[hashable]` you get a DataArray), `Sequence[Hashable]` -> select multiple variables (also from xarray, now if you do `dataset[sequence]` you get another dataset). 

If we forced users to interface directly with the xarray objects we wouldn't have to care about this, but the `var_names` and `coords` are two of the main arguments, especially `var_names` for which we allow excluding variables and fancier filtering which would be much harder to achieve with xarray. 

In general I think we want to maintain the xarray behaviour: None or sequence of hashable returns a dataset, hashable returns a dataarray. But there are cases like here where we do always want a dataset for internal use which won't be returned to the user. So we need to know what is valid in `dataset[xyz]` and what will be its result so we can adapt our code to these two cases as needed.

src/arviz_stats/psens.py
OriolAbril: I had missed the `.sum("latent-obs_var")` piece. So we sum over sample dims only. We have already summed over all variables and dims.

OriolAbril:
I have only looked at the `psens` function, not the `psens_summary` one. I think everything works as expected. Hopefully what is going on is somewhat clear. I think results shoud exactly match current arviz now, there was an issue in passing down the dimensions on which to operate dims->axes so instead of axes being defined from `dims` it was set to the default `-1` (instead of the correct `(-2, -1)` when using data with `chain, draw`).

Now both:

```python
import arviz_stats
from arviz_base import load_arviz_data, extract
idata = load_arviz_data("non_centered_eight")
arviz_stats.psens(idata, group="log_likelihood")
```

and 

```python
arviz_stats.psens(
    idata.drop_nodes(["prior", "prior_predictive", "observed_data", "constant_data"]).stack(sample=["chain", "draw"]),
    group="log_likelihood",
    sample_dims="sample"
)
```

work and return the same result.

src/arviz_stats/base/density.py
```diff
@@ -456,7 +457,7 @@ def kde_linear(
             x_min, x_max, x_std, extend_fct, grid_len, custom_lims, extend, bound_correction
         )
         grid_counts, grid_edges = self._histogram(
-            x, bins=grid_len, range=(grid_min, grid_max), density=False
+            x, bins=grid_len, weights=weights, range=(grid_min, grid_max), density=False
```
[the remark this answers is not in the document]
aloctavodia: I am reverting changes related to weighted kde. Those changes should be part of a separate PR.

src/arviz_stats/psense.py
```diff
@@ -0,0 +1,199 @@
...
+            return "strong prior / weak likelihood"
+
+        return "-"
+
+    psense_df["diagnosis"] = psense_df.apply(_diagnose, axis=1)
+
+    return psense_df.round(round_to)
+
+
+def _get_power_scale_weights(
```
OriolAbril, line 174: I prefer the original version. Or maybe have this function return both lower_w and upper_w?

Subsetting and summing over the "latent-obs_var" dimension should generally not be expensive computations, but they are not free and there is also no need to do that twice.

src/arviz_stats/psense.py
```diff
@@ -0,0 +1,199 @@
...
+        indices.append(labeller.make_label_flat(var_name, sel, isel))
+    psense_df.index = indices
+
+    def _diagnose(row):
+        if row["prior"] >= threshold and row["likelihood"] >= threshold:
+            return "prior-data conflict"
+        if row["prior"] > threshold > row["likelihood"]:
+            return "strong prior / weak likelihood"
+
+        return "-"
```
OriolAbril, line 167: I am wondering if we'd break anything by using `✓` here instead of a dash, afaik python supports unicode.

aloctavodia: Let's see!

src/arviz_stats/psense.py
OriolAbril: Only one of them? or both of them? If both we can modify the function to return both lower_w and upper_w. Otherwise it is probably better to have this be a `get_component_draws` function so only 2 lines are needed after that to get the weights

src/arviz_stats/psense.py
```diff
@@ -0,0 +1,193 @@
...
+    data : obj
+        Any object that can be converted to an :class:`arviz.InferenceData` object.
+        Refer to documentation of :func:`arviz.convert_to_dataset` for details.
+        For ndarray: shape = (chain, draw).
+        For n-dimensional ndarray transform first to dataset with ``az.convert_to_dataset``.
+    group : {"log_prior", "log_likelihood"}, default "log_prior"
+        When `component` is "likelihood", the log likelihood values are retrieved
+        from the ``log_likelihood`` group as pointwise log likelihood and added
+        together. With "prior", the log prior values are retrieved from the
+        ``log_prior`` group.
```
OriolAbril, line 45: Do we want `group` argument or `component` argument like in current ArviZ here? The docstring still mentions `component` argument but it is not used anymore.

OriolAbril (commented):
I think this is nearly ready to merge, only some api and doc related points left

src/arviz_stats/psense.py
aloctavodia: `group` sounds more consistent with other functions. But not opposed to have `component`.

OriolAbril: From the component api I like it is `prior`/`likelihood` without the log, a bit more conceptual. I think in general `group` is expected to potentially be any of the available groups everywhere except in `plot_ppc` where also similarly we use `posterior`/`prior` but use `posterior_predictive`/`prior_predictive` groups, so using prior/likelihood would align on this end here too. Between `component` or `group` as the argument name I don't really care, but I would update the `component_var_names` and coords if we use `group` (or at least document very clearly), as `group` will probably be more intuitive to users.

PR #34, opened 2024-10-24
Add more arguments to psense_summary

aloctavodia:
This also:

* Reorder arguments, matching the order in arviz-plots. i.e. first general arguments like var_names and coords and then specific ones. Is this what we want? Or we want first the specific ones? 
* Add `psense` and `psense_summary` to the API documentation

----
📚 Documentation preview 📚: https://arviz-stats--34.org.readthedocs.build/en/34/

### DIFF
--- a/docs/source/api/index.md
+++ b/docs/source/api/index.md
@@ -1,5 +1,15 @@
 # API reference
 
+## Functions
+
+```{eval-rst}
+.. autosummary::
+   :toctree: generated/
+
+   arviz_stats.psense.psense
+   arviz_stats.psense.psense_summary
+```
+
 ## Accessors
 Currently, using accessors is the recommended way to call functions from `arviz_stats`.
 
--- a/docs/source/api/index.md
+++ b/docs/source/api/index.md
@@ -6,8 +6,8 @@
 .. autosummary::
    :toctree: generated/
 
-   arviz_stats.psense.psense
-   arviz_stats.psense.psense_summary
+   arviz_stats.psense
+   arviz_stats.psense_summary
 ```
 
 ## Accessors
--- a/src/arviz_stats/psense.py
+++ b/src/arviz_stats/psense.py
@@ -24,7 +24,7 @@ def psense(
     group="prior",
     coords=None,
     sample_dims=None,
-    delta=0.01,
+    alphas=(0.99, 1.01),
     group_var_names=None,
     group_coords=None,
 ):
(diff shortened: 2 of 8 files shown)


docs/source/api/index.md
```diff
@@ -1,5 +1,15 @@
 # API reference
 
+## Functions
+
+```{eval-rst}
+.. autosummary::
+   :toctree: generated/
+
+   arviz_stats.psense.psense
+   arviz_stats.psense.psense_summary
```
OriolAbril, line 10:
```suggestion
   arviz_stats.psense
   arviz_stats.psense_summary
```

src/arviz_stats/psense.py
```diff
@@ -78,10 +77,15 @@ def psense(
     References
     ----------
     .. [1] Kallioinen et al, *Detecting and diagnosing prior and likelihood sensitivity with
-       power-scaling*, 2022, https://arxiv.org/abs/2107.14054
+    power-scaling*, Stat Comput 34, 57 (2024), https://doi.org/10.1007/s11222-023-10366-5
```
OriolAbril, line 80:
```suggestion
       power-scaling*, Stat Comput 34, 57 (2024), https://doi.org/10.1007/s11222-023-10366-5
```

The indentation to the `[x]` level is needed for correct rendering

src/arviz_stats/psense.py
```diff
@@ -106,15 +110,42 @@ def psense(
     )
 
 
-def psense_summary(data, threshold=0.05, round_to=3):
+def psense_summary(
```
OriolAbril, line 113: Should we support sample_dims here too?

OriolAbril (commented):
> first general arguments like var_names and coords and then specific ones. Is this what we want? Or we want first the specific ones?

Not completely sure. I think `var_names` and `coords` are the most commonly used arguments _in plots_, but I haven't really checked and also might not extend beyond plots.

The main advantage I see to common ones first is at some point I'd like to make all arguments beyond a cutoff (3rd or 4th) keyword only which would make things more intuitive (i.e. all signatures for plots start with `plot_xyz(idata, var_names, coords. *, other_args...)` so we could do `plot_xyz(idata, ["mu", "sigma"], {"group": ["a", "c"]}, ...)` independently of the plot. But it could also be argued that making the key specific arguments the only ones that could be positional might be more intuitive

aloctavodia:
Regarding the last commit. Priorsense doesn't use a delta value. Instead, it allows users to pass lower and upper values for alpha independently.

src/arviz_stats/psense.py
```diff
@@ -19,14 +19,14 @@
 
 def psense(
     dt,
+    var_names=None,
+    filter_vars=None,
     group="prior",
+    coords=None,
     sample_dims=None,
+    alphas=(0.99, 1.01),
```
OriolAbril, line 27: Isn't `alphas[0]` basically ignored now? Looks like I could do `(0.1, 1.01)` and `(0.99, 1.01)` and get the same result (maybe messing up the labels?)

aloctavodia: alpha[0] is not passed to `power_scale_sense` I think it should. I have reported this to the priorsense main dev. And he will check this ASAP.

PR #51, opened 2025-01-27
Add summary function

### LINKED ISSUE #11: Add summary
Add a summary function.

It might help to have a helper to "compile" multiple functions into a single ufunc with tuple output. But before diving too much into this a quick benchmark should be done to make sure it gives some advantage

aloctavodia:
Closes #11 

This adds the basic functionality for the `summary` function. It is still not on par with the `summary` function from old ArviZ, but it is better than nothing. ~Currently, the output is a dataset. To get a Pandas' DataFrame we need arviz-devs/arviz-base#25~

In the old arviz we had a `kind` argument and a `stat_focus` argument. The combination of both specifies which stats/diagnostics were computed. Now we achieve the same with just the `kind` argument. For instance `kind="all"` is the equivalent of the old `kind=all` and `stats_focus="mean"` and kind  `kind="all_median"` is the equivalent of the old `kind=all` and `stats_focus="median"`. Having a single argument seems clear to me.
 
One source of inconsistency is that previously we returned the `hdi` if `stat_focus="mean"` and the `eti` if `stat_focus="median` . But in arviz 1.0 whether to compute these two different CI depend on a global `rcparam` value. ~Should we ignore the `rcparam` specification and return what the `kind` argument dictates?~ If "stats_median", or "all_median" we return `eti` irrespective of the value of `ci_kind` (or the value in rcparams).

----
📚 Documentation preview 📚: https://arviz-stats--51.org.readthedocs.build/en/51/

### DIFF
--- a/src/arviz_stats/summary.py
+++ b/src/arviz_stats/summary.py
@@ -143,18 +143,18 @@ def summary(
 
     if kind in ["stats", "all"]:
         mean = dataset.mean(dim=sample_dims, skipna=skipna).expand_dims(summary=["mean"])
-        std = dataset.std(dim=sample_dims, skipna=skipna).expand_dims(summary=["std"])
+        std = dataset.std(dim=sample_dims, skipna=skipna).expand_dims(summary=["sd"])
         if ci_kind == "eti":
             ci = (
                 dataset.azstats.eti(prob=ci_prob, dims=sample_dims, skipna=skipna)
                 .rename({"quantile": "summary"})
-                .assign_coords(summary=[f"eti_{ci_perc}_lb", f"eti_{ci_perc}_ub"])
+                .assign_coords(summary=[f"eti{ci_perc}_", f"eti{ci_perc}^"])
             )
         else:
             ci = (
                 dataset.azstats.hdi(prob=ci_prob, dims=sample_dims, skipna=skipna)
                 .rename({"hdi": "summary"})
-                .assign_coords(summary=[f"hdi_{ci_perc}_lb", f"hdi_{ci_perc}_ub"])
+                .assign_coords(summary=[f"hdi{ci_perc}_", f"hdi{ci_perc}^"])
             )
         to_concat.extend((mean, std, ci))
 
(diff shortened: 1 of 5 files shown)


src/arviz_stats/summary.py
```diff
@@ -0,0 +1,222 @@
+"""Summaries for various statistics and diagnostics."""
+
+import xarray as xr
+from arviz_base import dataset_to_dataframe, extract, rcParams
+from xarray_einstats import stats
```
amaloney, line 5:
```suggestion
from typing import Hashbale, Literal, Sequence

import xarray as xr
from arviz_base import dataset_to_dataframe, extract, rcParams
from xarray.core.datatree import DataTree
from xarray_einstats import stats
```

src/arviz_stats/summary.py
```diff
@@ -0,0 +1,222 @@
...
+        For ndarray: shape = (chain, draw).
+        For n-dimensional ndarray transform first to dataset with ``az.convert_to_dataset``.
+    var_names : list of str, optional
+        Names of posterior variables to include in the power scaling sensitivity diagnostic
+    filter_vars: {None, "like", "regex"}, default None
+        Used for `var_names` only.
+        If ``None`` (default), interpret var_names as the real variables names.
+        If "like", interpret var_names as substrings of the real variables names.
+        If "regex", interpret var_names as regular expressions on the real variables names.
+    group : {"prior", "likelihood"}, default "prior"
```
amaloney, line 40: default in the function signature is `posterior`. I think there is a pre-commit directive we can use to ensure the function signature match the docstrings.

src/arviz_stats/summary.py
```diff
@@ -0,0 +1,222 @@
...
+        to_concat.extend((mean, std, ci))
+
+    if kind in ["diagnostics", "all"]:
+        ess_bulk = dataset.azstats.ess(dims=sample_dims, method="bulk").expand_dims(
+            summary=["ess_bulk"]
+        )
+        ess_tail = dataset.azstats.ess(dims=sample_dims, method="tail").expand_dims(
+            summary=["ess_tail"]
+        )
+        rhat = dataset.azstats.rhat(dims=sample_dims).expand_dims(summary=["R̂"])
```
amaloney, line 168: nice font!

amaloney (changes requested):
This looks great. Just a few nits about the docstring and type-hints, if you are okay with the suggested changes.

The only thing I would request an actual change for is the default value in the docstring does not match that in the function signature.

src/arviz_stats/summary.py
aloctavodia: that would be nice!

aloctavodia:
I fixed the docstring and added type hints. 

Please check if they are ok, feel free to make changes if you want. Not sure about the type hints for `data`, currently the function accepts DataSet, DataTree and InferenceData. Support for InferenceData should be transitional as in ArviZ 1.0 we replace it with DataTree.

amaloney (approved):
I think this is a great start. There are some new plumbings in Python 3.13 that will make more complicated types easier to handle, but I think what you have now will work just fine.

PR #84, opened 2025-03-24
Add loo_pit function

aloctavodia:
----
📚 Documentation preview 📚: https://arviz-stats--84.org.readthedocs.build/en/84/


sethaxen:
Comparing to `loo_pit` in `arviz`, this doesn't detect and smooth discrete data (see https://github.com/arviz-devs/arviz/blob/main/arviz/stats/stats.py#L1842-L1843)

Would that come in a later PR, or is that being discontinued?

PR #46, opened 2025-01-12
Add `bayes_factor` Function for Computations and Test Cases (#45)

PiyushPanwarFST:
This PR addresses issue [#45](https://github.com/arviz-devs/arviz-stats/issues/45) by refactoring and enhancing the Bayes Factor computation functionality. Key updates include:

## New Functionality

Added bayes_factor in psense.py for computing the Bayes Factor statistic, with optional plotting support.
Decoupled computation logic `az.bayes_factor` from visualisation `az.plot_bf` to ensure better modularity.

## Test Coverage

Added comprehensive tests in `test_psense.py` to validate functionality and cover edge cases.

## Code Quality

Documented the function using Numpy-style docstrings.
Ensured proper formatting with black and verified code quality using pylint and lint.sh.

## Benefits

Clear separation of concerns between computation `az.bayes_factor` and visualisation `az.plot_bf`.
Enhanced focus on statistical computations in arviz-stats.
Simplified dependencies for users needing computation-only logic.

## Additional Suggestion

I propose adding _kde_linear functionality to the density.py file, as implemented in the main ArviZ repository. This addition would further improve the modularity and functionality of  `arviz-stats`. I plan to raise a separate issue for this enhancement.

----
📚 Documentation preview 📚: https://arviz-stats--46.org.readthedocs.build/en/46/

PR #95, opened 2025-04-14
Update KDE computation & Bayes factor implementation (#55)

### LINKED ISSUE #55: Inplement bayes factor in a more ArviZian way

PiyushPanwarFST:
* As you suggested [Link](https://github.com/arviz-devs/arviz-stats/pull/46#issuecomment-2601783660) on Jan 20, this PR replaces the `_DensityBase` instantiation logic with the KDE accessor for density computation.

* Avoided the use of `_DensityBase` for KDE computation.

* Added `.azstats.kde()` accessor to compute densities and grids directly.

* Updated Bayes factor calculation to use the new KDE accessor.

* Simplified posterior/prior extraction using idata.posterior[var_name].

* Closes #55 

----
📚 Documentation preview 📚: https://arviz-stats--95.org.readthedocs.build/en/95/

### DIFF
--- a/src/arviz_stats/bayes_factor.py
+++ b/src/arviz_stats/bayes_factor.py
@@ -3,14 +3,14 @@
 import warnings
 
 import numpy as np
-from arviz_base import extract
-
-from arviz_stats.base.density import _DensityBase
 
 
 def bayes_factor(idata, var_name, ref_val=0, return_ref_vals=False, prior=None):
     """
-    Approximated Bayes Factor for comparing hypothesis of two nested models.
+    Compute Bayes factor using Savage–Dickey ratio.
+
+    We can apply this method when the null (`ref_val`) is a particular valueof the model
+    we are building. See [1]_ for details.
 
     Parameters
     ----------
@@ -42,12 +42,12 @@ def bayes_factor(idata, var_name, ref_val=0, return_ref_vals=False, prior=None):
         In [1]: import numpy as np
            ...: from arviz_base import from_dict
            ...: import arviz_stats as azs
-           ...: idata = from_dict({"posterior":{"a":np.random.normal(1, 0.5, (2, 1000))},
-           ...:                    {"prior":{"a":np.random.normal(0, 1, (2, 1000))}}})
-           ...: azs.plot_bf(idata, var_name="a", ref_val=0)
+           ...: dt = from_dict({"posterior":{"a":np.random.normal(1, 0.5, (2, 1000))},
+           ...:                    "prior":{"a":np.random.normal(0, 1, (2, 1000))}})
+           ...: azs.bayes_factor(dt, var_name="a", ref_val=0)
     """
-    posterior = extract(idata, var_names=var_name).values
-    prior = extract(idata, var_names=var_name, group="prior").values
+    posterior = idata.posterior[var_name]
+    prior = idata.prior[var_name]
 
     if not isinstance(ref_val, int | float):
         raise ValueError("The reference value (ref_val) must be a numerical value (int or float).")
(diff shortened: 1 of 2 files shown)


aloctavodia (changes requested):
While here make two changes

* Add this line to the init file from arviz_stats.bayes_factor import bayes_factor

* Change the docstring:
    """
    "Compute Bayes factor using Savage–Dickey ratio.
 
     We can apply this method when the null (`ref_val`) is a particular value of the model   we are building see [1]_ for details.

    Parameters
    ----------

aloctavodia:
I was thinking about a more general solution allowing Bayes factor function to work on many variables, not just one (and probably using interp from xarray). Still, this is a step in the right direction. So I am going to accept it (after changes) and then open a separate issue.

For future reference, https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/using-keywords-in-issues-and-pull-requests

src/arviz_stats/bayes_factor.py
```diff
@@ -3,14 +3,14 @@
...
-
-from arviz_stats.base.density import _DensityBase
 
 
 def bayes_factor(idata, var_name, ref_val=0, return_ref_vals=False, prior=None):
     """
-    Approximated Bayes Factor for comparing hypothesis of two nested models.
+    Compute Bayes factor using Savage–Dickey ratio.
+
+    We can apply this method when the null (`ref_val`) is a particular valueof the model
```
aloctavodia, line 12:
```suggestion
    We can apply this method when the null (`ref_val`) is a particular value of the model
```

src/arviz_stats/bayes_factor.py
```diff
@@ -46,8 +46,8 @@ def bayes_factor(idata, var_name, ref_val=0, return_ref_vals=False, prior=None):
            ...:                    {"prior":{"a":np.random.normal(0, 1, (2, 1000))}}})
```
aloctavodia, line 46:
```suggestion
           ...:                    {"prior":{"a":np.random.normal(0, 1, (2, 1000))}})
```

src/arviz_stats/bayes_factor.py
```diff
@@ -43,11 +43,11 @@ def bayes_factor(idata, var_name, ref_val=0, return_ref_vals=False, prior=None):
            ...: from arviz_base import from_dict
            ...: import arviz_stats as azs
            ...: idata = from_dict({"posterior":{"a":np.random.normal(1, 0.5, (2, 1000))},
-           ...:                    {"prior":{"a":np.random.normal(0, 1, (2, 1000))}}})
+           ...:                    {"prior":{"a":np.random.normal(0, 1, (2, 1000))}})
```
aloctavodia, line 46:
```suggestion
           ...:                    "prior":{"a":np.random.normal(0, 1, (2, 1000))}})
```

src/arviz_stats/bayes_factor.py
```diff
@@ -43,11 +43,11 @@ def bayes_factor(idata, var_name, ref_val=0, return_ref_vals=False, prior=None):
            ...: from arviz_base import from_dict
            ...: import arviz_stats as azs
            ...: idata = from_dict({"posterior":{"a":np.random.normal(1, 0.5, (2, 1000))},
-           ...:                    {"prior":{"a":np.random.normal(0, 1, (2, 1000))}}})
+           ...:                    "prior":{"a":np.random.normal(0, 1, (2, 1000))}})
```
aloctavodia, line 46: The example should be 

import numpy as np
from arviz_base import from_dict
import arviz_stats as azs
dt = from_dict({"posterior":{"a":np.random.normal(1, 0.5, (2, 1000))},
                    "prior":{"a":np.random.normal(0, 1, (2, 1000))}})
azs.bayes_factor(dt, var_name="a", ref_val=0)

PR #101, opened 2025-04-22
Add loo_approximate_posterior function

jordandeklerk:
This PR adds a new function, `loo_approximate_posterior`, that extends PSIS-LOO-CV to approximate posteriors from variational inference and other methods.

----
📚 Documentation preview 📚: https://arviz-stats--101.org.readthedocs.build/en/101/

### DIFF
--- a/src/arviz_stats/loo.py
+++ b/src/arviz_stats/loo.py
@@ -596,33 +596,8 @@ def loo_approximate_posterior(data, log_p, log_q, pointwise=None, var_name=None)
         [log_likelihood[dim].size for dim in log_likelihood.dims if dim not in sample_dims]
     )
 
-    for name, log_dens in [("log_p", log_p), ("log_q", log_q)]:
-        if isinstance(log_dens, np.ndarray):
-            if len(log_dens) != n_samples:
-                raise ValueError(
-                    f"Length of {name} ({len(log_dens)}) must match "
-                    f"the total number of samples in log_likelihood ({n_samples})."
-                )
-            log_dens_values = log_dens.reshape(log_likelihood.chain.size, log_likelihood.draw.size)
-            if name == "log_p":
-                log_p = xr.DataArray(
-                    log_dens_values,
-                    dims=sample_dims,
-                    coords={dim: log_likelihood[dim] for dim in sample_dims},
-                )
-            else:
-                log_q = xr.DataArray(
-                    log_dens_values,
-                    dims=sample_dims,
-                    coords={dim: log_likelihood[dim] for dim in sample_dims},
-                )
-        elif not isinstance(log_dens, xr.DataArray):
-            raise TypeError(f"{name} must be a numpy.ndarray or xarray.DataArray")
-
-    for dim in sample_dims:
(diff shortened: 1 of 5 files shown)


src/arviz_stats/loo.py
```diff
@@ -483,6 +484,202 @@ def loo_pit(data, var_names=None, log_weights=None, randomize=False):
...
+    # Handle underflow/overflow
+    log_ratio_max = corrected_log_ratios[var_name].max(dim=sample_dims)
+    corrected_log_ratios[var_name] = corrected_log_ratios[var_name] - log_ratio_max
+
+    # ignore r_eff here, set to r_eff=1.0
+    log_weights, pareto_k = corrected_log_ratios.azstats.psislw(r_eff=1.0, dims=sample_dims)
+    pareto_k_da = pareto_k[var_name]
+    log_weights += log_likelihood
+
+    warn_mg = False
```
aloctavodia, line 627: All of this code is shared with the `loo` function. We should put all this inside a function that we then call from both `loo` and `loo_approximate_posterior`

src/arviz_stats/loo.py
```diff
@@ -483,6 +484,202 @@ def loo_pit(data, var_names=None, log_weights=None, randomize=False):
...
+        - **n_samples**: number of samples
+        - **n_data_points**: number of data points
+        - **warning**: True if the estimated shape parameter of Pareto distribution is greater
+          than ``good_k``.
+        - **elpd_i**: :class:`~xarray.DataArray` with the pointwise predictive accuracy, only if
+          ``pointwise=True``
+        - **pareto_k**: array of Pareto shape values, only if ``pointwise=True``
+        - **good_k**: For a sample size S, the threshold is computed as
+          ``min(1 - 1/log10(S), 0.7)``
+
```
aloctavodia, line 533: Add entry for "approx_posterior", idem in the docstring of `loo`

src/arviz_stats/loo.py
```diff
@@ -483,6 +484,202 @@ def loo_pit(data, var_names=None, log_weights=None, randomize=False):
...
+           ...: values_q = rng.normal(loc=-1, scale=1, size=(log_lik.chain.size, log_lik.draw.size))
+           ...: log_q = xr.DataArray(
+           ...:     values_q,
+           ...:     dims=["chain", "draw"],
+           ...:     coords={"chain": log_lik.chain, "draw": log_lik.draw}
+           ...: )
+
+    This example focuses on demonstrating the function's usage. The provided approximate
+    posterior is deliberately simple and may not accurately represent the data, potentially
+    leading to less meaningful LOO results.
```
aloctavodia, line 565: I think we should move this text before the code. It could be ok to leave a sentence here to break the code into dummy posterior creation first and using function second.

src/arviz_stats/loo.py
jordandeklerk: Created `_compute_loo_results` to handle all of the shared warning messages and ELPD computations.

src/arviz_stats/loo.py
jordandeklerk: Moved this to the top of the example. I was conflicted as to where to place this. I think this makes sense.

PR #106, opened 2025-04-29
Add loo_subsample function for large datasets

jordandeklerk:
Adds two new capabilities that make PSIS-LOO-CV more practical on large data sets:

1. Adds `loo_subsample`, which performs leave-one-out cross-validation on a representative subset of observations to facilitate computations for large data.

2. Adds `update_subsample`, a companion function that can enhance an existing sub-sample result by adding additional observations.

To support these features, lower-level validation, data-extraction, and computation code has been moved into a new helper module called `helper_loo.py`.  That module relies on `namedtuple` objects to keep the various LOO components clearly organized, which in turn keeps the main functions easier to read and maintain.  A corresponding test file, `test_helper_loo.py`, verifies the functionality of the LOO helpers.

----
📚 Documentation preview 📚: https://arviz-stats--106.org.readthedocs.build/en/106/


src/arviz_stats/helper_loo.py
```diff
@@ -0,0 +1,561 @@
...
+
+    if param_names is None:
+        param_keys = sorted(list(posterior.data_vars.keys()))
+    else:
+        missing_params = [p for p in param_names if p not in posterior.data_vars]
+        if missing_params:
+            raise ValueError(f"Parameters {missing_params} not found in posterior group.")
+        param_keys = param_names
+
+    param_means = [float(posterior[key].mean()) for key in param_keys]
```
OriolAbril, line 118: I haven't had time to read the paper on this approximation so I hope I am not too far gone, but I think the documentation need extra details (in the `loo_subsample` `log_lik_fn` argument as it is the user facing side of this and probably an example too. The main thing that strikes me as odd and would make me second guess myself multiple times as a user is writing the log lik fn for scalars.

Now using eight schools as example as it is also the test case. However, then we take the global mean (here) so even though for the eight schools for example, `theta` which is the `mu` parameter of the Normal likelihood is a vector and each observation gets a different mean, here they all get the same mean instead of the respective observation mean, but we use `theta.mean()` instead of `mu.mean()` which is the parameter that represents the global mean.

---

The loo functions don't follow the separation of array interface first then dataarray/idata interface yet, but I do like writing the function as the scalar version as it can then be passed downstream and everything should work.

src/arviz_stats/helper_loo.py
```diff
@@ -0,0 +1,561 @@
...
+    plpd_values = np.empty_like(observed.values, dtype=float)
+
+    for idx, value in np.ndenumerate(observed.values):
+        try:
+            result = log_lik_fn(value, *param_means)
+            if not log:
+                result = np.log(np.maximum(result, np.finfo(float).tiny))
+
+            if not isinstance(result, int | float | np.number):
```
OriolAbril, line 127: NumPy recommends using `isinstance(value, numbers.Number)` (see for example [isscalar](https://numpy.org/doc/2.2/reference/generated/numpy.isscalar.html) docstring). Most code in old ArviZ does other things like this one but I think we should change the pattern for new code.

src/arviz_stats/helper_loo.py
```diff
@@ -0,0 +1,561 @@
...
+                raise TypeError(
+                    f"log_lik_fn must return a numeric scalar. Got type: {type(result)}"
+                )
+            plpd_values[idx] = result
+        except Exception as e:
+            raise RuntimeError(
+                f"Error computing log-likelihood with log_lik_fn for data point "
+                f"at index {idx} (value: {value}): {e}"
```
OriolAbril, line 135: Given the input here was an xarray object, showing the positional index might not mean much to the user. I would show the coordinate values instead. For example, given the dataarray `observed` we could do the following:

```python
observed[idx].coords
# output is multiline, something like
# Coordinates:
#     chain    int64 8B 0
#     draw     int64 8B 30
#     school   <U16 64B 'Phillips Exeter'
```

src/arviz_stats/helper_loo.py
```diff
@@ -0,0 +1,561 @@
...
+    tuple[float, float, float]
+        A tuple containing:
+
+        - elpd_loo_hat: The estimated ELPD using the difference estimator.
+        - subsampling_se_val: The standard error due to subsampling uncertainty.
+        - se_val: The total standard error (approximation + sampling uncertainty).
+    """
+    pointwise_diff = elpd_loo_i_sample - lpd_approx_sample
+    lpd_approx_sum_all = lpd_approx_all.sum().values
+    scaled_mean_pointwise_diff = n_data_points * pointwise_diff.mean().values
```
OriolAbril, line 180: we might want `.item()` here to get a scalar out instead of a 0d array.

src/arviz_stats/helper_loo.py
```diff
@@ -0,0 +1,561 @@
...
+
+
+def _prepare_loo_inputs(data, var_name):
+    """Prepare inputs for PSIS-LOO-CV."""
+    data = convert_to_datatree(data)
+
+    log_likelihood_ds = get_log_likelihood_dataset(data, var_names=var_name)
+    if var_name is None:
+        var_name = list(log_likelihood_ds.data_vars.keys())[0]
+    log_likelihood = log_likelihood_ds[var_name]
```
OriolAbril, line 298: I think this should use `get_log_likelihood` instead of the dataset version, then if var_name is none it can be `log_lik_da.name`. As the dataarray will come from being a variable in a dataset it must be named.

src/arviz_stats/helper_loo.py
```diff
@@ -0,0 +1,561 @@
...
+
+    sample_dims = ["chain", "draw"]
```
OriolAbril, line 300: not necessarily in this PR but then we should open an issue. This should use the rcParam and work for data where there is only a `sample` dimension for example because the user has stacked the dims manually or used `az.extract` on it.

OriolAbril (approved):
I left some small comments, none are issues for merging, several are not even introduced in this PR but wanted to point them out. Can't wait to try it

.pylintrc
```diff
@@ -134,7 +134,7 @@ indent-string='    '
 max-line-length=100
 
 # Maximum number of lines in a module
-max-module-lines=1000
+max-module-lines=2000
```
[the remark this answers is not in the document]
jordandeklerk: The issue I'm having is that the `loo.py` module is currently at almost 1500 lines with the helper functions moved into `helper_loo.py`. I'm not sure if we want to keep adding additional loo-based functions into the main `loo.py` file, in which case, the file will easily be larger than 2000 lines with additions like `loo_moment_match` (soon to come). 

Happy to re-organize things though.

src/arviz_stats/helper_loo.py
jordandeklerk: This is something I've been thinking a lot about as well. Not exactly sure the best approach forward, but I think we can keep iterating on this. I updated the documentation and added an example. Let me know what you think about it. Happy to add/modify for better clarity.

src/arviz_stats/helper_loo.py
jordandeklerk: Fixed this

src/arviz_stats/helper_loo.py
jordandeklerk: Added the following. Is this what you were thinking?

```python
        except Exception as e:
            coord_info = ", ".join(
                [f"{dim}: {observed[idx].coords[dim].values.item()}" for dim in observed.coords]
            )
            raise RuntimeError(
                f"Error computing log-likelihood with log_lik_fn for data point "
                f"with coordinates ({coord_info}) "
                f"at index {idx} (value: {value}): {e}"
            ) from e
```

src/arviz_stats/helper_loo.py
jordandeklerk: Fixed

src/arviz_stats/helper_loo.py
jordandeklerk: Great point, this is much cleaner now. Adjusted all the downstream usage for this as well.

src/arviz_stats/helper_loo.py
jordandeklerk: Agreed, I think this would be great.

jordandeklerk:
The newest commit fixes the changes mentioned by @OriolAbril and adds functionality for posterior thinning for `loo_subsample`. I'm using the `thin` function from `manipulation.py`. I had to add a thin method into `accessors.py` for DataArrays. 

Also storing the thinning factor in the `ELPD` object so that if the user calls `update_subsample`, we're using the same posterior draws as the original `loo_subsample` object.

.pylintrc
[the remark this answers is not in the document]
OriolAbril: Looked at the wrong files to check the lines, disregard

src/arviz_stats/helper_loo.py
OriolAbril: Splitting things into different layers like we do for `ess` for example should help as there are for example ufunc constructors that take number of dims to reduce and handle things for the function to work independently of those being 1, 2 or 4, but loo is significanly more complex so this machinery might need some updating.

src/arviz_stats/helper_loo.py
OriolAbril: We can keep iterating yes, in addition to the explanatory content on the stats side in https://arviz-devs.github.io/EABM/Chapters/Model_comparison.html we can also add notebooks as how-to resource for different situations here later on

jordandeklerk:
One last quick fix to add a new helper function to organize ELPD and Pareto value preparation. Ready to merge if all looks good @OriolAbril.

OriolAbril (changes requested):
Left one comment on the thin accessor that needs to be addressed. Other than that it is good to go, maybe we could also move the thin tests somewhere else but not completely sure where, I'll try to check if there already are some tests on `thin` somewhere if I have time.

src/arviz_stats/accessors.py
```diff
@@ -166,6 +166,45 @@ def _apply(self, func, **kwargs):
             func = get_function(func)
         return func(self._obj, **kwargs)
 
+    def thin(self, dims=None, factor="auto"):
```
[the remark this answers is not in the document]
jordandeklerk: So this was just my ignorance on how this worked with calling the DataArray functions from `accessors.py` as I was getting an error before, apologies for that. 

I just moified the `thin` function in `accessors.py` so I'm just calling the original thin function from `dataarray.py` now. Everything should be fine now since that function only thins across a single dimension and raises the error if the data can't be thinned. 

I can't find any other place with tests for thinning, so I'm inclined to leave the tests I have in `test_loo.py` and `test_helper_loo.py`. Happy to move those somewhere else though.

OriolAbril (approved):

OriolAbril:
I'll check the docs and merge

OriolAbril:
I think the doc build should be fixed now

PR #108, opened 2025-05-01
Add R2

aloctavodia:
Not sure where to place this. Once we add loo_R2 they could be together or loo_R2 should go inside loo.py. ~Also, not sure if we should break the Python variable name convention and call this function `R2`, or follow that rule and call it `r2` or something else~.

----
📚 Documentation preview 📚: https://arviz-stats--108.org.readthedocs.build/en/108/


sethaxen:
My preference would be to follow Python convention and go with `r2` instead of `R2`. Another good option would be `r2_score`, which scikit-learn uses.

OriolAbril (commented):
I think this would fit quite well into the array/dataarray/other layers.

Something like:

array_stats.r_squared(y_true, y_pred, axis) -> array_like with (sample, dims) shape
array_stats.r2_summary(y_true, y_pred, axis, point_estimate...) -> calls r2_score and returns 3 elem tuple or namedtuple

similar thing but dataarray inputs and defaults from rcParams
dataarray_stats.r_squared
dataarray_stats.r2_summary

arviz_stats.r2_score -> multiple options for inputs, maybe with the idata+name as default?

Doesn't need to happen in this PR though but wanted to comment and see what you think

src/arviz_stats/regression.py
```diff
@@ -0,0 +1,107 @@
...
+    y_true = extract(data, group="observed_data", var_names=obs_var_name, combined=False).values
+    y_pred = extract(data, group="posterior_predictive", var_names=pred_var_name).values.T
+
+    r_squared = array_stats.r2_score(y_true, y_pred)
+
+    if summary:
+        estimate = getattr(np, point_estimate)(r_squared).item()
+        c_i = getattr(array_stats, ci_kind)(r_squared, ci_prob)
+
+        r2_summary = namedtuple("R2", [point_estimate, f"{ci_kind}_lb", f"{ci_kind}_ub"])
```
aloctavodia, line 100: For `azs.summary`, we are using the notation f"{ci_kind}_", f"{ci_kind}_^". We can not do that for namedtuple as "^" is not supported. Whatever we use, it should be consistent.

Issue #86, opened 2025-03-25
loo_pit missing support for not matching variable names.


aloctavodia:
See #85

PR #115, opened 2025-05-08
Add support for different variable names in loo_pit

jordandeklerk:
Support for mis-matched variable names in `loo_pit` and corresponding test in `test_loo.py`.

resolves: [86](https://github.com/arviz-devs/arviz-stats/issues/86)

----
📚 Documentation preview 📚: https://arviz-stats--115.org.readthedocs.build/en/115/

### DIFF
--- a/src/arviz_stats/loo.py
+++ b/src/arviz_stats/loo.py
@@ -20,6 +20,7 @@ from arviz_stats.helper_loo import (
     _prepare_loo_inputs,
     _prepare_subsample,
     _prepare_update_subsample,
+    _process_var_name,
     _select_obs_by_coords,
     _srs_estimator,
     _warn_pareto_k,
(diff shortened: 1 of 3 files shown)


src/arviz_stats/loo.py
```diff
@@ -340,14 +341,21 @@ def loo_metrics(data, kind="rmse", var_name=None, round_to="2g"):
...
 
 
-def loo_pit(data, var_names=None, log_weights=None, randomize=False):
+def loo_pit(
+    data,
+    var_names=None,
+    log_weights=None,
+    randomize=False,
+    observed_var_names=None,
+    pp_var_names=None,
```
aloctavodia, line 350: In many functions in arviz_stats and arviz_plots we have the argument `data_pairs` , I think we want to keep that consistent.

    data_pairs : dict, optional
        Dictionary of keys prior/posterior predictive data and values observed data variable names.
        If None, it will assume that the observed data and the predictive data have
        the same variable name.

jordandeklerk: Good point. I overlooked this!

jordandeklerk:
Latest commit adds the `data_pairs` argument to `loo_pit` and also fixes a normalization issue. Before, the `loo_pit` values in the example were the following:

```python
from arviz_stats import loo_pit
from arviz_base import load_arviz_data, from_dict
dt = load_arviz_data("centered_eight")
loo_pit(dt)

<xarray.Dataset> Size: 576B
Dimensions:  (school: 8)
Coordinates:
  * school   (school) <U16 512B 'Choate' 'Deerfield' ... 'Mt. Hermon'
Data variables:
    obs      (school) float64 64B 170.9 777.6 1.337e+03 ... 298.9 721.7
```

Clearly, too large and not between 0 and 1. The following normalization fix seems to have corrected this:

```python
    loo_pit_values = np.exp(
        logsumexp(log_weights.where(pit_vals, -np.inf), dims=["chain", "draw"])
        - logsumexp(log_weights, dims=["chain", "draw"])
    )
```

I think before we were summing the weighted counts where the condition was met without dividing by the total. Hopefully I'm not missing anything.

aloctavodia (approved):

PR #119, opened 2025-05-16
Add wasserstein distance and KL-divergence

### LINKED ISSUE #118: Support for distance metrics
Would distance metrics like wasserstein and kl-divergence (ok ok it's not a proper distance) be within the scope of this project? It would be nice to have an out-of-the-box way to compute these on pairs of arviz posteriors

aloctavodia:
Closes #118 

----
📚 Documentation preview 📚: https://arviz-stats--119.org.readthedocs.build/en/119/

### DIFF
--- a/metrics.py
+++ b/metrics.py
@@ -1,0 +1,102 @@
+"""Collection of metrics for evaluating the performance of probabilistic models."""
+
+from arviz_base import convert_to_datatree, dataset_to_dataframe, extract, rcParams
+from scipy.stats import wasserstein_distance, wasserstein_distance_nd
+
+
+def wasserstein(
+    data_1,
+    data_2,
+    group="posterior",
+    var_names=None,
+    sample_dims=None,
+    joint=True,
+    num_samples=500,
+    random_seed=212480,
+):
+    """
+    Compute the Wasserstein-1 distance.
+
+    Parameters
+    ----------
+    data : Dataset, DataTree, or InferenceData
+    group : hashable, default "posterior"
+        Group on which to compute the ESS.
+    var_names : str or list of str, optional
+        Names of the variables for which the ess should be computed.
+    sample_dims : iterable of hashable, optional
+        Dimensions to be considered sample dimensions and are to be reduced.
+        Default ``rcParams["data.sample_dims"]``.
+    joint : bool, default True
+        If True, compute the joint Wasserstein distance. If False, compute the
+        Wasserstein distance for each marginal. The computation is faster when `joint=False`
+        but it is equivalent to assume that the marginals are independent, which usually
+        is not the case.
+    num_samples : int
+        Number of samples to use for the Wasserstein distance calculation. Default is 500.
+    random_seed : int
+        Random seed for reproducibility. Use None for no seed.
+
+    Returns
+    -------
+    wasserstein_distance : float
+    """
+    data_1 = convert_to_datatree(data_1)
+    data_2 = convert_to_datatree(data_2)
+    if sample_dims is None:
+        sample_dims = rcParams["data.sample_dims"]
+
+    dist1 = _extract_and_reindex(
+        data_1,
+        group=group,
+        var_names=var_names,
+        sample_dims=sample_dims,
+        num_samples=num_samples,
+        random_seed=random_seed,
+    )
+
+    dist2 = _extract_and_reindex(
+        data_2,
+        group=group,
+        var_names=var_names,
+        sample_dims=sample_dims,
+        num_samples=num_samples,
+        random_seed=random_seed,
+    )
+
+    shared_var_names = set(dist1.data_vars) & set(dist2.data_vars)
+    if not shared_var_names:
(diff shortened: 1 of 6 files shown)


metrics.py
```diff
@@ -0,0 +1,102 @@
...
+            data,
+            group=group,
+            var_names=var_names,
+            num_samples=num_samples,
+            random_seed=random_seed,
+            keep_dataset=True,
+        )
+        .reset_index("sample")
+        .drop_vars(sample_dims)
+        .assign_coords(sample=range(num_samples))
```
OriolAbril, line 101: I think we need to improve the `dataset_to_dataarray` to be able to handle this. I am still not completely groking the explicit indexes and the differences with multiindex but this case is common enough that I think we need to handle extract+to_dataarray/dataframe works as is.

metrics.py
```diff
@@ -0,0 +1,102 @@
...
+    shared_var_names = set(dist1.data_vars) & set(dist2.data_vars)
+    if not shared_var_names:
+        raise ValueError("No shared variable names found.")
+
+    if var_names is None:
+        var_names = list(shared_var_names)
+        dist1, dist2 = dist1[var_names], dist2[var_names]
+
+    dist1 = dataset_to_dataframe(dist1, sample_dims=["sample"]).values
+    dist2 = dataset_to_dataframe(dist2, sample_dims=["sample"]).values
```
OriolAbril, line 76: I didn't think about using extract first, that is very nice. And as extract has already stacked all `sample_dims` we can then use "only" dataset_to_dataarray, no need to do the extra step to dataframe as the output will already be 2d array.

metrics.py
```diff
@@ -0,0 +1,102 @@
...
+    var_names : str or list of str, optional
+        Names of the variables for which the ess should be computed.
+    sample_dims : iterable of hashable, optional
+        Dimensions to be considered sample dimensions and are to be reduced.
+        Default ``rcParams["data.sample_dims"]``.
+    joint : bool, default True
+        If True, compute the joint Wasserstein distance. If False, compute the
+        Wasserstein distance for each marginal. The computation is faster when `joint=False`
+        but it is equivalent to assume that the marginals are independent, which usually
+        is not the case.
```
OriolAbril, line 34: I think I would move the last sentence to the [notes](https://numpydoc.readthedocs.io/en/latest/format.html#notes) section and there also mention that this happens by calling either one or the other scipy functions with links to their docs.

metrics.py
```diff
@@ -0,0 +1,102 @@
...
+    joint=True,
+    num_samples=500,
+    random_seed=212480,
+):
+    """
+    Compute the Wasserstein-1 distance.
+
+    Parameters
+    ----------
+    data : Dataset, DataTree, or InferenceData
```
OriolAbril, line 22:
```suggestion
    data_1, data_2 : Dataset, DataTree, or InferenceData
```

numpydoc allows parameters that have the same type and similar description to be documented in the same line. Ref: https://numpydoc.readthedocs.io/en/latest/format.html#parameters

> When two or more input parameters have exactly the same type, shape and description, they can be combined:
>
>     x1, x2 : array_like
>         Input arrays, description of `x1`, `x2`.

metrics.py
OriolAbril, line 1: I have comented on this file and now seen it is not the right one as it is not part of the package

src/arviz_stats/metrics.py
```diff
@@ -105,3 +107,270 @@ def r2_score(
...
+
+def _extract_and_reindex(data, group, var_names, sample_dims, num_samples, random_seed):
+    return (
+        extract(
+            data,
+            group=group,
+            var_names=var_names,
+            num_samples=num_samples,
+            random_seed=random_seed,
+            keep_dataset=True,
```
OriolAbril, line 317: sample_dims needs to be passed to extract too

src/arviz_stats/metrics.py
```diff
@@ -105,3 +107,270 @@ def r2_score(
...
+
+    Compute KL-divergence using equation 14 from [1]_. Assumes both arrays
+    are of the same shape.
+
+    Parameters
+    ----------
+    ary0 : array-like
+        First sample of shape (n_posterior_samples, n_outputs).
+    ary1 : array-like
+        Second sample of shape (n_posterior_samples, n_outputs).
```
OriolAbril, line 336:
```suggestion
    ary0, ary1 : (N, M) array-like
        Samples of the input distributions. ``N`` represents the number of samples (e.g. posterior samples)
        and ``M`` the number of outputs (e.g. number of variables in the posterior)
```

It is common to include the shape information in the type, and being same type and shape we can also share type and description between the arguments

OriolAbril (approved):
I left some minor comments we can merge after they have been addressed, I also opened https://github.com/arviz-devs/arviz-base/issues/55 to fix things in arviz-base, once fixed we can come back here and simplify the code.

metrics.py
aloctavodia: agreed

PR #120, opened 2025-05-16
Fix loo pit randomization

aloctavodia:
The randomisation routine was cumbersome and contained errors. This is a more straightforward implementation and closer to what is done in posterior package. Not sure this is the best one from an xarray perspective. The test `test_loo_pit_different_names` is failing I think the problem is that the dimensions are not aligned when doing 

        sel_min[var] = posterior_predictive[pp_var_names[i]] < observed_data[var]
        sel_sup[var] = posterior_predictive[pp_var_names[i]] == observed_data[var]

I will not have the time to look at this during the weekend, so if someone wants to clean this, go ahead. Otherwise, I can continue with this next week.

----
📚 Documentation preview 📚: https://arviz-stats--120.org.readthedocs.build/en/120/

### DIFF
--- a/src/arviz_stats/loo.py
+++ b/src/arviz_stats/loo.py
@@ -343,9 +343,7 @@ def loo_metrics(data, kind="rmse", var_name=None, round_to="2g"):
 def loo_pit(
     data,
     var_names=None,
-    data_pairs=None,
     log_weights=None,
-    randomize=False,
 ):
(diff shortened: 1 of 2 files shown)


aloctavodia:
Great, thanks!

aloctavodia:
This code works, but the implementation does not seem to be very idiomatic from an array perspective @OriolAbril may you have a better idea how to approach this?

PR #121, opened 2025-05-17
Add "regular" metrics

aloctavodia:
This adds a series of metrics mirroring those already available in `loo_metrics`. To ease comparison between predictive metrics and this one, they are defined in the same way. 

~I guess for a future PR we may want to move the computation to the array interface, similar to what we already have for r2_score~.

----
📚 Documentation preview 📚: https://arviz-stats--121.org.readthedocs.build/en/121/

### DIFF
--- a/src/arviz_stats/base/diagnostics.py
+++ b/src/arviz_stats/base/diagnostics.py
@@ -760,3 +760,118 @@ class _DiagnosticsBase(_CoreBase):
         var_e = np.var(y_true - y_pred, axis=1)
         r_squared = var_y_est / (var_y_est + var_e)
         return r_squared
+
+    @staticmethod
+    def _mae(observed, predicted):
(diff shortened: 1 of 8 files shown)


src/arviz_stats/base/diagnostics.py
```diff
@@ -760,3 +760,138 @@ def _r2_score(y_true, y_pred):
...
+        predicted: array-like of shape = (n_outputs)
+            Estimated target values.
+
+        Returns
+        -------
+        tuple (mean, std_error)
+            mean: float
+                Mean absolute error.
+            std_error: float
+                Standard error of the mean absolute error.
```
OriolAbril, line 781:
```suggestion
        mean: float
            Mean absolute error.
        std_error: float
            Standard error of the mean absolute error.
```

these can be directly documented like this

OriolAbril (approved):

PR #122, opened 2025-05-17
Dim fix

jordandeklerk:
Fixes the dimension error from `test_loo_pit_different_names`. I think the issue is that when we use `data_pairs` to map between different variable names, the resulting PIT values form a matrix where each element $p_{ij}$ represents $P(pred_{j} \leq y_i | y_{-i})$. For LOO-PIT, I think we only need the diagonal elements.

### DIFF
--- a/src/arviz_stats/loo.py
+++ b/src/arviz_stats/loo.py
@@ -483,9 +483,16 @@ def loo_pit(
             )
 
             pit_upper = pit_lower + pit_sup_addition[var].values
+
+            if pit_lower.ndim > 1:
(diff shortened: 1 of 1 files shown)


jordandeklerk:
In my head, this latest commit here looked good. But I'm now realizing this is very similar to the last commit in [#120](https://github.com/arviz-devs/arviz-stats/pull/120). I guess I'm not sure what the best approach is in handling the naming mismatch without explicitly renaming the pred var names to match the observation names.

aloctavodia:
Based on the discussion over Slack, we can close this in favour of #120 .

PR #123, opened 2025-05-20
add ci_in_rope

aloctavodia:
Compute the percentage of a credible interval that falls within a ROPE.


src/arviz_stats/summary.py
```diff
@@ -215,3 +216,129 @@ def summary(
...
+    elif isinstance(rope, dict):
+        rope_low = xr.full_like(ci_low, fill_value=np.nan)
+        rope_high = xr.full_like(ci_high, fill_value=np.nan)
+
+        for var, (low, high) in rope.items():
+            if var in rope_low:
+                rope_low[var] = low
+                rope_high[var] = high
+    else:
+        raise ValueError("`rope` must be a tuple or dict")
```
OriolAbril, line 337: I think this could be handled by `references_to_dataset` with maybe an extra check to make sure all variables in input data are present in rope dataset?

src/arviz_stats/summary.py
```diff
@@ -215,3 +216,129 @@ def summary(
...
+        ci_kind = rcParams["stats.ci_kind"] if "stats.ci_kind" in rcParams else "eti"
+    if sample_dims is None:
+        sample_dims = rcParams["data.sample_dims"]
+
+    dataset = extract(
+        data,
+        var_names=var_names,
+        filter_vars=filter_vars,
+        group=group,
+        combined=False,
```
OriolAbril, line 309: I would set combined=True and then use `dims="sample"` in hdi/eti

now thinking it might need to be `dims=sample_dims if isinstance(sample_dims, str) else "sample"

src/arviz_stats/summary.py
```diff
@@ -215,3 +216,129 @@ def summary(
...
+
+    if ci_kind == "eti":
+        c_i = dataset.azstats.eti(prob=ci_prob, dims=sample_dims).rename({"quantile": "ci_kind"})
+    else:
+        c_i = dataset.azstats.hdi(prob=ci_prob, dims=sample_dims).rename({"hdi": "ci_kind"})
+
+    stacked = dataset.stack(sample=sample_dims)
+
+    ci_low = c_i.sel(ci_kind="lower")
+    ci_high = c_i.sel(ci_kind="higher")
```
OriolAbril, line 321: I am not sure this will work, but this also raises a potentially interesting question. Do we want to change the dims/coords of hdi/eti output to match?

aloctavodia: This only works because I did a rename earlier. Yes, they should match, that will simplify things for us and potentially for users too.

src/arviz_stats/summary.py
aloctavodia: Not sure how to do this. I tried a couple of alternatives before setting on this. Do you want to try?

src/arviz_stats/summary.py
```diff
@@ -215,3 +218,136 @@ def summary(
...
+            raise ValueError(f"{rope_dim} is not a dimension of `rope`")
+        if (rope_len := rope.sizes[rope_dim]) != 2:
+            raise ValueError(f"Length of {rope_dim} dim must be 2 but is {rope_len}")
+    else:
+        try:
+            rope = np.array(rope)
+        except ValueError as err:
+            raise ValueError(
+                "`rope` must be a dict, Dataset or array-like, failed to convert to array"
+            ) from err
```
OriolAbril, line 332: I have no problem restricting to tuple, but I really think dataset should also be a valid input

OriolAbril (commented):
I din't break the tests but might be worth doing some user testing anyway, also ended taking care of hdi/eti renaming here to `ci_bound`, we can use only `bound` too but was afraid it might be too general

src/arviz_stats/summary.py
aloctavodia: ok

aloctavodia:
"ci_bound" sounds good

Issue #63, opened 2022-10-12
[Feature] Implement importance weighted moment matching


n-kall:
## Description

Importance weighted moment matching (IWMM) is an implicitly adaptive importance sampling method that improves proposal distributions in importance sampling by performing affine transformations on the proposal draws. This can be used in e.g. loo-cv or power-scaling sensitivity analysis, when Pareto-smoothed importance sampling is not stable.

The paper describing the method is available here: https://link.springer.com/article/10.1007/s11222-020-09982-2 and a reference method is implemented in R: https://github.com/topipa/iwmm
The loo R package (https://github.com/stan-dev/loo) also has an implementation.

## Thoughts on implementation

In order to run IWMM, I think the following are required:
- unconstrained parameter draws
- method to compute the posterior density for specified parameter values
- method to transform unconstrained draws to constrained draws

I'm happy to help with this, as I have been working with the R code and translating it should not be too hard, however I'm unsure of the best approach to get access to the above requirements (as they aren't usually in an InferenceData object)

aloctavodia:
Yes, this will be implemented in this repo (I just transfered the issue)

mlpotter:
@aloctavodia that is great to hear. Is there an expected timeline / release time for this functionality?

aloctavodia:
I hope within the next ~4 months.

jordandeklerk:
Working on this now. Hoping to have a solid initial implementation of this by end of this month!

PR #126, opened 2025-05-26
Add loo_moment_match

jordandeklerk:
This PR adds `loo_moment_match`,  implementing the moment matching algorithm for PSIS-LOO-CV described in [Paananen et al. (2021)](https://arxiv.org/pdf/1906.08850). This implementation addresses the limitation of standard PSIS-LOO-CV when many Pareto k values are high (k > 0.7), avoiding exact model refitting. The moment matching approach iteratively adjusts posterior draws to better approximate the leave-one-out posterior distribution.

### Major Additions

- **`loo_moment_match()`**: Main function handling all the logic and main moment matching algorithm. 

- **Affine Transformations**: 
  - `_shift()`: Simple mean-matching transformation
  - `_shift_and_scale()`: Matches both mean and marginal variances
  - `_shift_and_cov()`: Full covariance matching (when `cov=True`)

- **Split transformation**: `_split_moment_match` gives the option to transform only half the draws and use multiple importance sampling for improved stability

- **Helper functions**: Support for recalculating importance weights, updating pointwise PSIS-LOO-CV data, etc.

### Notes

- Had to expand the line limit for invidual modules to 2500 since the `loo.py` module is getting quite big. Maybe we want to move things around?
- Included `pymc` as a dependency for building the documentation. I wanted to include a realistic example using a PPL for moment matching to give the user an idea of how to define and use the required user-provided functions, `log_prob_upars_fn` and `log_lik_i_upars_fn`. Happy to modify this. 
- In [63](https://github.com/arviz-devs/arviz-stats/issues/63), it's mentioned that we might need a function which maps the adjusted, unconstrained posterior draws back to the constrained space, e.g.., a constraining bijector. To my understanding, this is not a requirement in [loo_moment_matching.R](https://github.com/stan-dev/loo/blob/master/R/loo_moment_matching.R), which is the main source of inspiration for this PR. This functionality does exist in [iwmm](https://github.com/topipa/iwmm/tree/main), however. Happy to discuss adding this here as well.

---

Resolves [63](https://github.com/arviz-devs/arviz-stats/issues/63)

### DIFF
--- a/docs/source/api/index.md
+++ b/docs/source/api/index.md
@@ -62,6 +62,7 @@ you should jump to {ref}`array_stats_api` and read forward.
    arviz_stats.loo_approximate_posterior
    arviz_stats.loo_subsample
    arviz_stats.update_subsample
+   arviz_stats.loo_moment_match
 ```
 
 ### Other
--- a/tests/test_loo.py
+++ b/tests/test_loo.py
@@ -533,6 +533,7 @@ def test_loo_moment_match(radon_problematic, loo_orig, moment_match_data):
         split=True,
         max_iters=50,
     )
+
     assert isinstance(loo_mm, ELPDData)
     assert loo_mm.kind == "loo"
     assert loo_mm.method == "loo_moment_match"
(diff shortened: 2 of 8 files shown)


aloctavodia:
* We can move things around, maybe having a loo module with one function per file or grouped somehow. 
* I would keep the example relatively simple. For more elaborate examples, we have https://arviz-devs.github.io/EABM/, Currently, we have two "chapters" about model comparison https://arviz-devs.github.io/EABM/Chapters/Model_comparison.html and https://arviz-devs.github.io/EABM/Chapters/Case_study_model_comparison.html We can expand one or both chapters or even add a new one. Then we can add a link to the docs of `loo_moment_match` pointing to these extended resources.

jordandeklerk:
- Added a new `loo` module with separate files for each loo-related function.
- Simplified the `loo_moment_match` example
- Removed the PyMC dependency 
- Reverted module line length back to `max=1000`

docs/source/api/index.md
```diff
@@ -63,6 +63,7 @@ you should jump to {ref}`array_stats_api` and read forward.
    arviz_stats.loo_approximate_posterior
    arviz_stats.loo_subsample
    arviz_stats.update_subsample
+   arviz_stats.loo_moment_match
```
aloctavodia, line 66: order alphabetically

tests/test_loo.py
```diff
@@ -462,3 +468,156 @@ def test_update_subsample_preserves_thin_factor(centered_eight):
...
+        log_lik_i_upars_fn=log_lik_i_fn,
+        k_threshold=k_threshold,
+        var_name="y",
+        cov=True,
+        split=True,
+        max_iters=50,
+    )
+
+    assert isinstance(loo_mm, ELPDData)
+    assert loo_mm.kind == "loo"
```
aloctavodia, line 538: No need to fix here, but this makes me realize that we no longer need this attribute, as we only support loo, right?

src/arviz_stats/loo/loo_moment_match.py
```diff
@@ -0,0 +1,792 @@
...
+    loo_orig,
+    upars,
+    log_prob_upars_fn,
+    log_lik_i_upars_fn,
+    max_iters=30,
+    k_threshold=None,
+    split=True,
+    cov=False,
+    pointwise=None,
+    var_name=None,
```
aloctavodia, line 39: not sure how strict we are about this here (we are more consistent in arviz-plots), but usually var_name/var_names is listed among the first few arguments

aloctavodia (commented):
nitpicks

aloctavodia:
Have you compared this with R loo? Do you have an example we can add here https://arviz-devs.github.io/EABM/? If not, I can add it, just let me know. Maybe we can reproduce this vignette https://mc-stan.org/loo/articles/loo2-moment-matching.html

tests/test_loo.py
jordandeklerk: Right, good point. Since we aren't including WAIC, this isn't necessary anymore.

jordandeklerk:
> Have you compared this with R loo? Do you have an example we can add here https://arviz-devs.github.io/EABM/? If not, I can add it, just let me know. Maybe we can reproduce this vignette https://mc-stan.org/loo/articles/loo2-moment-matching.html

I haven't tested this yet against R loo, but I was planning on creating a tutorial using PyMC that re-creates that roaches vignette. I will add an issue for that and work on that. 

I think we could also add a tutorial for `loo_subsample` as well that can incorporate `loo_approximate_posterior` too. We could re-create this vignette https://mc-stan.org/loo/articles/loo2-large-data.html essentially.

aloctavodia:
LGTM! Let's see what we find when trying to reproduce examples.

PR #151, opened 2025-06-25
Add method for multiple chains, remove optimization method

aloctavodia:
I am adding the method for the envelope that takes into account multiple chains. I am implementing the cdf of the hypergeometric using numba, as the one in scipy is too slow for this. I visually compared the results with the example in bayesplot, and there is no clear difference. Results are also visually equivalent when using the custom function or the one from scipy. I added a small test comparing the computation in scipy and the custom cdf. 

I am removing the optimization method as it is currently slower than the simulation method, and the simulation method seems to be fast enough so far. We can later revisit this.

One more thing, I need to check in detail, but I think the default in bayesplot for the ci_prob when computing the envelope is 0.99, maybe the 0.94  will lead to too many "rejections". Should we have this default too. Maybe a new RcParams? ()

----
📚 Documentation preview 📚: https://arviz-stats--151.org.readthedocs.build/en/151/

### DIFF
--- a/src/arviz_stats/ecdf_utils.py
+++ b/src/arviz_stats/ecdf_utils.py
@@ -1,18 +1,21 @@
 """Utility functions for computing the PIT ECDF and the simultaneous confidence bands."""
 
 import math
+import warnings
 
 import numpy as np
 from arviz_base import dict_to_dataset
-from numba import vectorize
-from scipy.optimize import minimize_scalar
 from scipy.special import bdtr, bdtrik  # pylint: disable=no-name-in-module
-from scipy.stats import uniform
+
+try:
+    from numba import vectorize
+
+    NUMBA_FLAG = True
+except ImportError:
+    NUMBA_FLAG = False
 
 
-def difference_ecdf_pit(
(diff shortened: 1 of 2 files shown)


OriolAbril:
I have tried vectorizing the computation to avoid the loops over simulation and chain "dimensions" so the scipy distribution is only initialized once. Each hypergeom evaluation still took ~40 seconds on the data I was using (while took 1-2 seconds). This felt way too much for any C++ code, so I double checked the env and the c++ library mentioned in the scipy docs is an optional dependency which is not pulled in even by `conda install scipy` from conda forge. After installing all boost related libraries from conda forge the result was exactly the same.

It is disapointing but it also means the issue is not with initializing the distributions n_simulations*n_chains times, it is the function itself that is unbearably slow. I vote we merge for now and I'll open issues to: handle the backend defaults, what is included in conda default install, and restructure this into the computational module layout. I think in general we should install the library with numba but allow disabling it/not installing it (potentially showing warnings from extremely problematic cases like this one).

OriolAbril (approved):

PR #162, opened 2025-07-23
Organize computations for visualizations and make it easy to apply on groupbys

OriolAbril:
I have started with a helper function to handle the processing and calling of the
functions/accessors on the different potential inputs.

I would also like to reorganize the ecdf functions in the the class structure to have
inheritance and make numba an optional dependency (instead of required).

In addition, I will start using that on arviz-plots side to allow for groupby behaviour in
plot_ppc_dist for example.

### DIFF
--- a/src/arviz_stats/__init__.py
+++ b/src/arviz_stats/__init__.py
@@ -23,6 +23,7 @@ try:
     from arviz_stats.summary import summary, ci_in_rope
     from arviz_stats.manipulation import thin
     from arviz_stats.bayes_factor import bayes_factor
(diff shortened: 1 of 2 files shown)


OriolAbril:
I have moved the tests and docstrings from #168 here. I will add the docstrings for kde/ecdf/hist so we can merge. I will try and restructure ecdf in a different PR so I don't block this for too long

OriolAbril:
Still far from ideal but it should work. We can merge and continue improving docs and ecdf functionality later on

PR #174, opened 2025-08-12
add mode computation

### LINKED ISSUE #9: Add mode
Add mode computation.

aloctavodia:
closes #9 

For the mode computation I use the half-sample method, instead of the max density from a kde. According to what I read and a few tests it seems to provide reasonable results (even for circular variable without doing any change). It only returns a single mode.

This PR assumes we want the mode computation to behave like xarray's mean (or similar functions). So

```python
dt = azb.load_arviz_data("centered_eight").posterior
dt.azstats.mode()
``` 

returns 3 values, one for mu, theta, and tau. 

And 

```python
dt.azstats.mode(dim=["chain", "draw"])
``` 

returns 10 values, one for mu and tau, and eight for theta

This currently fails 

```python
dt.azstats.mode(dim=["school"])
``` 

If we were computing the mean, the result would be 3 arrays of shape (4, 500), one per variable.

----
📚 Documentation preview 📚: https://arviz-stats--174.org.readthedocs.build/en/174/

### DIFF
--- a/src/arviz_stats/base/dataarray.py
+++ b/src/arviz_stats/base/dataarray.py
@@ -402,5 +402,19 @@ class BaseDataArray:
             output_core_dims=[dims],
         )
 
+    def mode(self, da, dim=None):
+        """Compute mode on DataArray input."""
+        if dim is None:
+            dims = list(da.dims)
+        else:
+            dims = dim if isinstance(dim, (list | tuple)) else [dim]
+
+        return apply_ufunc(
+            self.array_class.mode,
+            da,
+            input_core_dims=[dims],
+            kwargs={"axis": np.arange(-len(dims), 0, 1)},
+        )
+
 
 dataarray_stats = BaseDataArray(array_class=array_stats)
--- a/src/arviz_stats/base/dataarray.py
+++ b/src/arviz_stats/base/dataarray.py
@@ -404,15 +404,13 @@ class BaseDataArray:
 
     def mode(self, da, dim=None):
         """Compute mode on DataArray input."""
-        if dim is None:
-            dims = list(da.dims)
-        else:
-            dims = dim if isinstance(dim, (list | tuple)) else [dim]
+        dims = validate_dims(dim)
 
         return apply_ufunc(
             self.array_class.mode,
             da,
             input_core_dims=[dims],
+            output_core_dims=[[]],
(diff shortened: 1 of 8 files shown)


src/arviz_stats/base/dataarray.py
```diff
@@ -402,5 +402,19 @@ def autocorr(self, da, dim=None):
             output_core_dims=[dims],
         )
 
+    def mode(self, da, dim=None):
```
aloctavodia, line 405: I guess this is not entirely correct, given the reported error

src/arviz_stats/summary.py
```diff
@@ -346,3 +347,103 @@ def ci_in_rope(
...
+
+    Returns
+    -------
+    ndarray, DataArray, Dataset, DataTree
+        Requested mode of the provided input.
+
+    See Also
+    --------
+    :func:`xarray.DataArray.mean`, :func:`xarray.Dataset.mean`, :func:`xarray.DataArray.median`,
+    :func:`xarray.Dataset.median`
```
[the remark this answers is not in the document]
OriolAbril: I think that is ok, we could also choose to link only to dataarray/dataset methods to keep things shorter if we prefer.

src/arviz_stats/base/dataarray.py
```diff
@@ -402,5 +402,19 @@ def autocorr(self, da, dim=None):
...
+        """Compute mode on DataArray input."""
+        if dim is None:
+            dims = list(da.dims)
+        else:
+            dims = dim if isinstance(dim, (list | tuple)) else [dim]
```
OriolAbril, line 410: I am not completely sure we want this as the behaviour. I see the appeal of matching the behaviour of `mean` and `median`, but at the same time we are also making this be the odd one out within `azstats` accessors and I am not sure where to draw the line. Why does `eti` behave differently than quantile?

disclaimer: I have a conflict of interest because I like arviz default much more than xarray one, so if this is the default I will be constantly writing `dim=["chain", "draw"]` and I get intrusive thoughts every few months about adding a `ds.azstats.mean()` that only does `ds.mean(rcParams["sample_dims"])`

aloctavodia: It could be a good idea to have an arviz's version of mean and median, but stop there! haha.

OriolAbril (approved):

Issue #181, opened 2025-08-24
Add loo_i function for single observation PSIS-LOO-CV testing


jordandeklerk:
We should implement a `loo_i()` function that computes LOO for a single observation, similar to what's available in the R loo package. 

Currently, users have to run the entire LOO computation even if they just want to check a single point. With `loo_i()`, they could quickly validate their log-likelihood functions and get detailed diagnostics for individual observations. Testing log-likelihood functions would be quite useful before running the full PSIS-LOO-CV computations for methods that require user-defined log-likelihood functions. 

The function would follow the same parameter conventions as our existing `loo()` function but operate on just a single observation. This would keep the API consistent and familiar for users who are already using the LOO functionality in arviz-stats.

PR #182, opened 2025-08-24
Add `loo_i` function for single observation PSIS-LOO-CV computation

jordandeklerk:
This adds the `loo_i()` function to compute PSIS-LOO-CV for a single observation. 

---

Resolves [#181](https://github.com/arviz-devs/arviz-stats/issues/181)

### DIFF
--- a/src/arviz_stats/loo/loo.py
+++ b/src/arviz_stats/loo/loo.py
@@ -1,12 +1,17 @@
 """Pareto-smoothed importance sampling LOO (PSIS-LOO-CV)."""
 
+import numpy as np
 from arviz_base import rcParams
+from xarray_einstats.stats import logsumexp
 
 from arviz_stats.loo.helper_loo import (
     _compute_loo_results,
+    _get_log_likelihood_i,
     _get_r_eff,
     _prepare_loo_inputs,
+    _warn_pareto_k,
 )
+from arviz_stats.utils import ELPDData
 
 
 def loo(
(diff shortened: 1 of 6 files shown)


src/arviz_stats/loo/loo.py
```diff
@@ -162,3 +167,151 @@ def loo(
...
+    .. [2] Vehtari et al. *Pareto Smoothed Importance Sampling*.
+       Journal of Machine Learning Research, 25(72) (2024) https://jmlr.org/papers/v25/19-556.html
+       arXiv preprint https://arxiv.org/abs/1507.02646
+    """
+    if not isinstance(i, int | np.integer):
+        raise TypeError(f"i must be an integer, got {type(i)}")
+
+    loo_inputs = _prepare_loo_inputs(data, var_name)
+
+    i = int(i)
```
aloctavodia, line 273: You checked  that i is integer

src/arviz_stats/loo/loo.py
```diff
@@ -162,3 +167,151 @@ def loo(
...
+):
+    r"""Compute PSIS-LOO-CV for a single observation.
+
+    Estimates the expected log pointwise predictive density (elpd) using Pareto-smoothed
+    importance sampling leave-one-out cross-validation (PSIS-LOO-CV) for a single observation.
+    The method is described in [1]_ and [2]_.
+
+    Parameters
+    ----------
+    i : int
```
aloctavodia, line 188: Maybe use idx?

aloctavodia (approved):

src/arviz_stats/loo/loo.py
jordandeklerk: This is a good point. I initially thought that this function just be for quick testing mainly with custom log-likelihood functions so the actual indexing wouldn't be relevant. But I think maybe we should account for more calculated options here for the user. 

The most recent commit allows for several different ways of doing this. Maybe this is overkill though?

PR #192, opened 2025-09-04
Rework PLPD and LPD approximation for `loo_subsample()`

jordandeklerk:


src/arviz_stats/loo/loo_subsample.py
```diff
@@ -86,26 +86,40 @@ def loo_subsample(
     method: str, optional
         Method used for approximating the pointwise log predictive density:
 
-        - 'lpd': Use standard log predictive density approximation (default)
-        - 'plpd': Use Point Log Predictive Density approximation which requires a ``log_lik_fn``.
+        - ``lpd``: Use standard log predictive density approximation (default)
+        - ``plpd``: Use point log predictive density approximation which requires a ``log_lik_fn``.
     thin: int, optional
```
aloctavodia, line 91: `thin` also supports "auto", any reason to only allow integers here?

aloctavodia (approved):

src/arviz_stats/loo/loo_subsample.py
jordandeklerk: Updated the docstring for this. The downstream helper function that prepares the `loo_subsample()` data handles this and accepts `str` for thinning, it just wasn't specified in the docs.

PR #216, opened 2025-10-01
Add subsampling support in `compare()`

jordandeklerk:
This adds model comparison support for subsampled LOO results, matching the behavior of the `loo` package.

When using `loo_subsample()`, the previous `compare()` returned NaN values for the difference standard error (`dse`) and didn't report a `subsample_dse` when using subsampled models. Subsampled results store NaN for non-sampled observations, and the pointwise variance calculation `np.var(diff)` propagated these NaNs which caused issues with `dse`. The `loo` package handles this by storing subsample metadata, which we track in functions for subsampling in `helper_loo.py`, and using a difference-of-estimators approach for variance calculation.

Also adds the same warnings from the `loo` package when comparing full LOO to subsampled LOO, and when comparing subsampled LOO models with different subsamples (defaults to naive SE in this case). 

----
📚 Documentation preview 📚: https://arviz-stats--216.org.readthedocs.build/en/216/

### DIFF
--- a/src/arviz_stats/loo/compare.py
+++ b/src/arviz_stats/loo/compare.py
@@ -90,6 +90,26 @@ def compare(
            ...: compare_dict = {"non centered": data1, "centered": data2}
            ...: compare(compare_dict)
 
+    Compare models using subsampled LOO:
+
+    .. ipython::  python
+        :okwarning:
+
+        In [1]: from arviz_stats import loo_subsample
+           ...: from arviz_base import load_arviz_data
+           ...: data1 = load_arviz_data("non_centered_eight")
+           ...: data2 = load_arviz_data("centered_eight")
+           ...: loo_sub1 = loo_subsample(data1, observations=6, pointwise=True, seed=42)
+           ...: loo_sub2 = loo_subsample(data2, observations=6, pointwise=True, seed=42)
+           ...: compare({"non_centered": loo_sub1, "centered": loo_sub2})
+
+    When using subsampled LOO, the ``subsampling_dse`` column quantifies the additional
+    uncertainty from using subsamples instead of all observations. The ``elpd_diff`` values
+    are computed using a difference-of-estimators approach on overlapping observations, which
+    can differ from simple subtraction of ELPD values. Using the same seed across models
+    ensures overlapping observations for more accurate paired comparisons with smaller
+    standard errors.
+
     See Also
     --------
     :func:`loo` : Compute the ELPD using the Pareto smoothed importance sampling Leave-one-out
--- a/src/arviz_stats/loo/compare.py
+++ b/src/arviz_stats/loo/compare.py
@@ -101,7 +101,7 @@ def compare(
            ...: data2 = load_arviz_data("centered_eight")
            ...: loo_sub1 = loo_subsample(data1, observations=6, pointwise=True, seed=42)
            ...: loo_sub2 = loo_subsample(data2, observations=6, pointwise=True, seed=42)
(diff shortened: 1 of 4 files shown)


src/arviz_stats/loo/compare.py
```diff
@@ -85,6 +90,26 @@ def compare(
...
+    .. ipython::  python
+        :okwarning:
+
+        In [1]: from arviz_stats import loo_subsample
+           ...: from arviz_base import load_arviz_data
+           ...: data1 = load_arviz_data("non_centered_eight")
+           ...: data2 = load_arviz_data("centered_eight")
+           ...: loo_sub1 = loo_subsample(data1, observations=6, pointwise=True, seed=42)
+           ...: loo_sub2 = loo_subsample(data2, observations=6, pointwise=True, seed=42)
+           ...: compare({"non_centered": loo_sub1, "centered": loo_sub2})
```
aloctavodia, line 104: The dataframe does not render fully in the example; maybe we could print fewer decimals; the same applies to the previous example.

```suggestion
           ...: compare({"non_centered": loo_sub1, "centered": loo_sub2}).round(2)
```

src/arviz_stats/loo/compare.py
```diff
@@ -85,6 +90,26 @@ def compare(
...
+
+    When using subsampled LOO, the ``subsampling_dse`` column quantifies the additional
```
aloctavodia, line 106: What should I do with this "extra-uncertainty"? Should I add it to the "regular one"? If that's the case, why not add it directly and optionally print a warning message (if useful or just explain that the uncertainty will be added when subsampling)?  Should we also update plot_compare to check if this column exists and modify the plot accordingly?

jordandeklerk: i dont think the `subsampling_dse` can be added to the regular one. They are not independent components in `_diff_srs_estimator()`:

```python
total_variance_estimate = (lpd_approx_sq_sum_all + scaled_mean_sq_diff) - (
    1 / n_data_points
) * (
    scaled_mean_pointwise_diff**2
    - subsampling_variance  # subtracted here
    + 2 * lpd_approx_sum_all * elpd_loo_estimate
    - lpd_approx_sum_all**2
)
```

So the `- subsampling_variance` term is effectively added back as `+ subsampling_variance / N` to the total variance. It acts as a finite population correction and is not a separable component. So `dse` already incorporates a subsampling adjustment, and combining it with `subsampling_dse` as `sqrt(dse^2 + subsampling_dse^2)` would double count.

The main benefit I see of including `subsampling_dse` (at the risk of maybe some confusion) in `compare()` is that `subsampling_dse` provides diagnostic information specific to model comparison. Individual `loo_subsample()` outputs show whether you have enough observations in the subsample to estimate each model's ELPD confidently, they don't tell you whether you have enough observations to confidently compare models that were subsampled. The `subsampling_dse` in `compare()` addresses this directly. It shows whether the subsampling uncertainty in the difference is small enough relative to `elpd_diff` to make a reliable comparison. Also, when models use overlapping subsamples, `subsampling_dse` is computed from pointwise differences accounting for correlation, which isn't derivable from individual model uncertainties.

We could probably make a note in the documentation if necessary to clarify that `dse` already includes a subsampling adjustment and should not be combined with `subsampling_dse`.

For `plot_compare()`, I think we should probably keep using the "regular SE" since it already includes the subsampling SE. I don't think there's much to gain from the plot by looking at a nested error bar if we were to also include the `subsample_dse`, honestly.

aloctavodia (approved):
We can add some notes clarifying this in EABM

PR #217, opened 2025-10-06
Patch computations in `compare()` for subsampling

jordandeklerk:
This fixes some issues in how we compute ELPD differences for subsampled models. Previously, we had some minor problems that led to slight miscalculations in variance estimates. This new approach converts everything to numpy arrays and flattens before indexing to avoid any dimension issues.

We effectively need to  address three distinct cases: 

1. when neither model uses subsampling, we compute standard pointwise differences
2. when only one model uses subsampling, we compute on overlapping observations and combine uncertainties
3. when both models use subsampling, we use the difference-of-estimators approach on the intersection of subsamples, with a fallback to naive diff when subsamples don't overlap

We also add some warnings to the user for certain scenarios.

### DIFF
--- a/src/arviz_stats/loo/compare.py
+++ b/src/arviz_stats/loo/compare.py
@@ -6,7 +6,6 @@ from copy import deepcopy
 
 import numpy as np
 import pandas as pd
-import xarray as xr
 from arviz_base import rcParams
 from scipy.optimize import minimize
 from scipy.stats import dirichlet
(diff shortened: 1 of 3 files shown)


src/arviz_stats/loo/compare.py
```diff
@@ -270,90 +269,181 @@ def gradient(weights):
...
+        "elpd_diff": elpd_diff,
+        "se_diff": se_diff,
         "subsampling_dse": subsampling_dse,
     }
 
 
+def _compute_naive_diff(elpd_a, elpd_b):
+    """Compute naive ELPD difference using paired observations."""
+    elpd_diff = elpd_a.elpd - elpd_b.elpd
+    se_a = getattr(elpd_a, "se", 0.0) or 0.0
```
aloctavodia, line 433: should this just be `se_a = getattr(elpd_a, "se", 0.0)`?

aloctavodia (approved):

jordandeklerk:
Merging this to finish off EABM chapter.

Issue #218, opened 2025-10-09
Improve overall test coverage


jordandeklerk:
As mentioned in the monthly ArviZ meeting, we need to test more of the core functionality in arviz-stats. I feel pretty good about most of the functionality for `loo` based functions. I've been able to find some very nuanced bugs in several functions there. I would like to expand more edge case testing to other statistical functions within the arviz-stats ecosystem.

PR #220, opened 2025-10-14
Refactor and expand tests

jordandeklerk:
This significantly refactors and expands the test suite for arviz-stats. The goal here was to test core functionality and relevant edge cases for areas that either lacked test coverage or had very minimal exposure. I also moved all `loo` based testing into its own testing directory so we could expand on those tests and separate the modules out. 

While this PR primarily focuses on robustifying the test suite, there were several changes to source code as well along the way. Some of these are (hopefully) small improvements, while others came up from some edge-case tests.

### Source Changes

- **[`src/arviz_stats/accessors.py:276`](https://github.com/arviz-devs/arviz-stats/blob/improve-tests/src/arviz_stats/accessors.py#L276)**
   - Fixed typo in error message: changed `self._obs.name` to `self._obj.name`

- **`src/arviz_stats/numba/intervals.py`**
  - Fixed missing `return` statement in [`quantile()` function (line 31)](https://github.com/arviz-devs/arviz-stats/blob/improve-tests/src/arviz_stats/numba/intervals.py#L31)
  - Fixed in-place assignment in [`_quantile()` (line 21)](https://github.com/arviz-devs/arviz-stats/blob/improve-tests/src/arviz_stats/numba/intervals.py#L21): changed `result = ...` to `result[:] = ...`

- **`src/arviz_stats/numba/diagnostics.py`**
  - Fixed DataTree group access in [`rhat()` (line 121)](https://github.com/arviz-devs/arviz-stats/blob/improve-tests/src/arviz_stats/numba/diagnostics.py#L121): changed `ds[group]` to `ds[group].dataset`
  - Fixed DataTree group access in [`ess()` (line 286)](https://github.com/arviz-devs/arviz-stats/blob/improve-tests/src/arviz_stats/numba/diagnostics.py#L286): changed `ds[group]` to `ds[group].dataset`

- **[`src/arviz_stats/sampling_diagnostics.py:388`](https://github.com/arviz-devs/arviz-stats/blob/improve-tests/src/arviz_stats/sampling_diagnostics.py#L388)**
  - Fixed recursive call: changed `rhat` to `rhat_nested` in `data.map()` call

- **[`src/arviz_stats/base/dataarray.py:325-328`](https://github.com/arviz-devs/arviz-stats/blob/improve-tests/src/arviz_stats/base/dataarray.py#L325-L328)**
  - Added defensive fallback in thinning factor calculation to ensure `factor >= 1`
  - Prevents issues when ESS exceeds n_samples (probably unlikely in practice with real data, but is possible with synthetic data)

- **`src/arviz_stats/survival.py`** ([lines 25-27](https://github.com/arviz-devs/arviz-stats/blob/improve-tests/src/arviz_stats/survival.py#L25-L27), [lines 106-108](https://github.com/arviz-devs/arviz-stats/blob/improve-tests/src/arviz_stats/survival.py#L106-L108))
  - Improved `var_names` parameter in `kaplan_meier()` and `generate_survival_curves()`
  - Now accepts both `str` and `list[str]` and is not listed as optional anymore

- **`src/arviz_stats/base/dataarray.py`** ([line 115](https://github.com/arviz-devs/arviz-stats/blob/improve-tests/src/arviz_stats/base/dataarray.py#L115), [lines 125-129](https://github.com/arviz-devs/arviz-stats/blob/improve-tests/src/arviz_stats/base/dataarray.py#L125-L129))
  - Added `method` parameter to `rhat_nested()` with default `"rank"`

- **[`src/arviz_stats/sampling_diagnostics.py`](https://github.com/arviz-devs/arviz-stats/blob/improve-tests/src/arviz_stats/sampling_diagnostics.py#L378-L416)** (lines 378-416)
  - Propagated `method` parameter throughout all `rhat_nested()` calls
  - Ensures consistent behavior across DataArray, Dataset, and DataTree interfaces

- **[`src/arviz_stats/numba/diagnostics.py:69`](https://github.com/arviz-devs/arviz-stats/blob/improve-tests/src/arviz_stats/numba/diagnostics.py#L69)**
  - Switched from `raw_rearrange` to `rearrange` (was getting an error for raw_rearrange)

I think all fixes should be backward compatible. I know this is quite a large PR. Things escalated quickly to say the least 😄. 

---
Resolves (first pass at least): https://github.com/arviz-devs/arviz-stats/issues/218

----
📚 Documentation preview 📚: https://arviz-stats--220.org.readthedocs.build/en/220/

### DIFF
--- a/tests/base/test_array.py
+++ b/tests/base/test_array.py
@@ -1,6 +1,6 @@
 """Tests for array interface base functions."""
 
-# pylint: disable=no-member, redefined-outer-name, no-self-use, protected-access
+# pylint: disable=redefined-outer-name, no-self-use, protected-access
 import numpy as np
 import pytest
 from numpy.testing import assert_allclose, assert_array_equal
--- a/tests/loo/test_loo_expectations.py
+++ b/tests/loo/test_loo_expectations.py
@@ -1,5 +1,6 @@
-# pylint: disable=redefined-outer-name, unused-import, unused-argument
-# ruff: noqa: F811, F401
+"""Test expectations functions for PSIS-LOO-CV."""
+
+# pylint: disable=redefined-outer-name, unused-argument
 import numpy as np
 import pytest
 from numpy.testing import assert_allclose, assert_almost_equal, assert_array_equal
--- a/tests/loo/test_reloo.py
+++ b/tests/loo/test_reloo.py
@@ -1,5 +1,6 @@
+"""Test re-loo for PSIS-LOO-CV."""
+
 # pylint: disable=redefined-outer-name, no-self-use, unused-argument
-# ruff: noqa: F811
 from copy import deepcopy
 
 import numpy as np
--- a/tests/base/test_core.py
+++ b/tests/base/test_core.py
@@ -1,7 +1,6 @@
 """Tests for core base functions."""
 
-# ruff: noqa: E731
-# pylint: disable=no-member, redefined-outer-name, no-self-use, protected-access
+# pylint: disable=redefined-outer-name, no-self-use, protected-access
 import numpy as np
 import pytest
 from numpy.testing import assert_allclose, assert_array_almost_equal, assert_array_equal
--- a/tests/base/test_density.py
+++ b/tests/base/test_density.py
@@ -1,6 +1,6 @@
 """Tests for density estimation base functions."""
 
-# pylint: disable=no-member, redefined-outer-name, no-self-use, protected-access
+# pylint: disable=redefined-outer-name, no-self-use, protected-access
 import numpy as np
 import pytest
 from numpy.testing import assert_allclose
--- a/tests/base/test_stats.py
+++ b/tests/base/test_stats.py
@@ -1,4 +1,6 @@
-# pylint: disable=redefined-outer-name, no-member, unused-import
+"""Test stats functions."""
+
+# pylint: disable=redefined-outer-name
 import numpy as np
 import pytest
 from numpy.testing import assert_array_almost_equal, assert_array_equal
@@ -9,7 +11,7 @@ from ..helpers import importorskip
 azb = importorskip("arviz_base")
 xr = importorskip("xarray")
 
-import arviz_stats.accessors  # noqa: F401
+import arviz_stats.accessors  # noqa: F401  # pylint: disable=unused-import
 
 
 @pytest.fixture(scope="module")
--- a/tests/base/test_stats_utils.py
+++ b/tests/base/test_stats_utils.py
@@ -1,7 +1,7 @@
 """Tests for stats_utils."""
 
 # ruff: noqa: E731
-#  pylint: disable=no-member,unnecessary-lambda-assignment
+#  pylint: disable=unnecessary-lambda-assignment
 import numpy as np
 import pytest
 from numpy.testing import assert_array_almost_equal
--- a/tests/helpers.py
+++ b/tests/helpers.py
@@ -1,4 +1,3 @@
-# pylint: disable=redefined-outer-name
 """Test related helper functions."""
 
 import os
--- a/tests/loo/test_compare.py
+++ b/tests/loo/test_compare.py
@@ -1,5 +1,6 @@
-# pylint: disable=redefined-outer-name, unused-import, unused-argument
-# ruff: noqa: F811, F401
+"""Test compare function."""
+
+# pylint: disable=redefined-outer-name, unused-argument
 import copy
 
 import numpy as np
--- a/tests/loo/test_helper_kfold.py
+++ b/tests/loo/test_helper_kfold.py
@@ -1,5 +1,4 @@
-# pylint: disable=redefined-outer-name, unused-import
-# ruff: noqa: F811
+# pylint: disable=redefined-outer-name
 """Test helper functions for k-fold cross-validation."""
 
 import numpy as np
--- a/tests/loo/test_helper_loo.py
+++ b/tests/loo/test_helper_loo.py
@@ -1,5 +1,6 @@
-# pylint: disable=redefined-outer-name, unused-argument, unused-import
-# ruff: noqa: F811, F401
+"""Test helper functions for PSIS-LOO-CV."""
+
+# pylint: disable=redefined-outer-name, unused-argument
 import warnings
 
 import numpy as np
--- a/tests/loo/test_loo.py
+++ b/tests/loo/test_loo.py
@@ -1,5 +1,6 @@
-# pylint: disable=redefined-outer-name, unused-import, unused-argument
-# ruff: noqa: F811, F401
+"""Test PSIS-LOO-CV."""
+
+# pylint: disable=redefined-outer-name, unused-argument
 import numpy as np
 import pytest
 from numpy.testing import assert_allclose, assert_almost_equal, assert_array_equal
--- a/tests/loo/test_loo_ap.py
+++ b/tests/loo/test_loo_ap.py
@@ -1,5 +1,6 @@
-# pylint: disable=redefined-outer-name, unused-import, unused-argument
-# ruff: noqa: F811, F401
+"""Test PSIS-LOO-CV for approximate posteriors."""
+
+# pylint: disable=redefined-outer-name
 import numpy as np
 import pytest
 
--- a/tests/loo/test_loo_kfold.py
+++ b/tests/loo/test_loo_kfold.py
@@ -1,7 +1,6 @@
-# pylint: disable=redefined-outer-name, unused-import, too-many-function-args
-# ruff: noqa: F811
 """Test k-fold cross-validation."""
 
+# pylint: disable=redefined-outer-name, too-many-function-args
 import numpy as np
 import pytest
 from numpy.testing import assert_almost_equal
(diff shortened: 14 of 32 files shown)


tests/base/test_array.py
```diff
@@ -0,0 +1,467 @@
...
+
+
+@pytest.fixture
+def array_stats():
+    return BaseArray()
+
+
+class TestHelperFunctions:
+    def test_process_chain_none_with_none(self):
+        ary = np.random.randn(100, 50)
```
OriolAbril: these should switch to using the newer numpy random generators. Using `RandomState` instead of `Generator` is discouraged, calling `numpy.random.function` directly even more so. Ref: https://numpy.org/doc/stable/reference/random/legacy.html#functions-in-numpy-random

EDIT: in this particular function and a few more below where we basically care about the shape only we are probably better off using `np.empty` though.

tests/base/test_array.py
```diff
@@ -0,0 +1,467 @@
...
+        ary = rng.vonmises(0, 2, size=1000)
+        result = array_stats.hdi(ary, prob=0.9, circular=True)
+        assert result.shape == (2,)
+
+    @pytest.mark.parametrize("axis", [0, 1, -1])
+    def test_hdi_axis(self, array_stats, axis):
+        rng = np.random.default_rng(42)
+        ary = rng.normal(size=(10, 20, 30))
+        result = array_stats.hdi(ary, prob=0.9, axis=axis)
+        assert result.shape[-1] == 2
```
OriolAbril: I would also check the reduced dim is the correct one

tests/base/test_array.py
```diff
@@ -0,0 +1,467 @@
...
+        assert result.shape == ()
+        assert result > 0
+
+    def test_mcse_invalid_method(self, array_stats):
+        ary = np.random.randn(4, 100)
+        with pytest.raises(ValueError, match="must be one of"):
+            array_stats.mcse(ary, method="invalid")
+
+
+class TestParetoFunctions:
```
OriolAbril: I would also add tests for ndarray input, with both integer and list/array input for `axis` argument (with the exception of pareto_min_ss which takes chain_axis and draw_axis so will probably look closer to the cases above)

tests/base/test_array.py
```diff
@@ -0,0 +1,467 @@
...
+        result = array_stats.compute_ranks(x, relative=True)
+        assert np.all(result >= 0)
+        assert np.all(result <= 1)
+
+    @pytest.mark.parametrize("axis", [0, 1, -1])
+    def test_compute_ranks_axis(self, array_stats, axis):
+        rng = np.random.default_rng(42)
+        ary = rng.normal(size=(10, 20, 30))
+        result = array_stats.compute_ranks(ary, axis=axis)
+        assert result.ndim == ary.ndim
```
OriolAbril: given the shape is always preserved independently of what is passed as `axis` I would check the maximum value within the array which will depend on the length of the dimension indicated by `axis`

tests/base/test_array.py
```diff
@@ -0,0 +1,467 @@
...
+
+
+class TestBinning:
+    def test_get_bins_basic(self, array_stats):
+        rng = np.random.default_rng(42)
+        ary = rng.normal(size=(100,))
+        bins = array_stats.get_bins(ary)
+        assert bins.ndim == 1
+        assert len(bins) > 0
```
OriolAbril: given the test is seeded maybe we should check the length of `bins` and add an assert against that?

I also forgot the combination of algorightms/rules we use as "arviz auto method" but it might also be a good idea to check bins[0] and [-1] against minimum and max of `ary`. I think the checks will be min=bins[0] and then either max=bins[-1] or max<=bins[-1]

tests/base/test_array.py
```diff
@@ -0,0 +1,467 @@
...
+class TestKDE:
+    def test_kde_basic(self, array_stats):
+        rng = np.random.default_rng(42)
+        ary = rng.normal(size=(100,))
+        grid, pdf, bw = array_stats.kde(ary)
+        assert len(grid) == 512
+        assert len(pdf) == 512
+        assert bw > 0
+
+    @pytest.mark.parametrize("grid_len", [128, 256, 512, 1024])
```
OriolAbril: I don't think this will actually take very long, but given the test nature, I think the first two are enough to check what we care about

tests/base/test_array.py
```diff
@@ -0,0 +1,467 @@
...
+        assert len(pdf) == grid_len
+
+    def test_kde_circular(self, array_stats):
+        rng = np.random.default_rng(42)
+        ary = rng.vonmises(0, 2, size=100)
+        grid, pdf, bw = array_stats.kde(ary, circular=True)
+        assert len(grid) == 512
+        assert len(pdf) == 512
+        assert bw > 0
+
```
OriolAbril: here I would also add a test for behaviour of `axis`

OriolAbril (commented):
I have started going over the tests I will continue as it will take a while. So far the most recurring comment is adding tests of `axis` behaviour in `array` functions. The xarray related tests are definitely passing the argument to the array function, so it might not have any effect on actual coverage if looked at from a global perspective.

However, there are already many functions with tests for `axis` and given we have a `minimal` test job that runs tests on an env without xarray nor arviz-base I think it would be nice to try and get the global coverage of the library and the coverage of minimal test job (counted with respect to things inside `base/*` except the dataarray file).

Second common comment is try to avoid `np.random.function` for new code and use the `default_rng()` then `rng.function`

All source code related changes look good.

tests/base/test_array.py
```diff
@@ -0,0 +1,467 @@
...
+        ary = rng.normal(size=(100,))
+        result = array_stats.get_bins(ary, bins=bins)
+        assert result.ndim == 1
+
+    def test_histogram_basic(self, array_stats):
+        rng = np.random.default_rng(42)
+        ary = rng.normal(size=(1000,))
+        counts, edges = array_stats.histogram(ary)
+        assert len(counts) == len(edges) - 1
+        assert counts.sum() == 1000
```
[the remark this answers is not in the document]
jordandeklerk: I can probably handle this in another quick PR.

jordandeklerk:
> I have started going over the tests I will continue as it will take a while. So far the most recurring comment is adding tests of `axis` behaviour in `array` functions. The xarray related tests are definitely passing the argument to the array function, so it might not have any effect on actual coverage if looked at from a global perspective.
> 
> However, there are already many functions with tests for `axis` and given we have a `minimal` test job that runs tests on an env without xarray nor arviz-base I think it would be nice to try and get the global coverage of the library and the coverage of minimal test job (counted with respect to things inside `base/*` except the dataarray file).
> 
> Second common comment is try to avoid `np.random.function` for new code and use the `default_rng()` then `rng.function`
> 
> All source code related changes look good.

Think I have all of the changes related to axes testing you mentioned, and fixed the `np.random.function` issues. Thanks for bringing that up, I always forget about this. 

Appreciate the review so far as well. Going to continue looking for places to test axis behavior where it makes sense. I feel pretty good about the `loo` tests in terms of coverage and core testing functionality. For functions or modules with no coverage (something like `survival.py`, for example), I tried to just keep it simple for now and was hoping for feedback on anything specific we wanted to test.

tests/loo/test_loo_expectations.py
```diff
@@ -0,0 +1,216 @@
...
+    if kind == "quantile":
+        assert loo_exp_vals.shape == (2, 8)
+    else:
+        assert loo_exp_vals.shape == (8,)
+
+    assert_almost_equal(loo_exp_vals.sel({"school": "Choate"}), expected_vals, decimal=2)
+
+
+@pytest.mark.filterwarnings("ignore::UserWarning")
+@pytest.mark.filterwarnings("ignore::RuntimeWarning")
```
OriolAbril: not the first I have seem but these are great, thanks! For legacy arviz we missed some warnings because the tests were too noisy and didn't see some deprecations until the function was removed for example

Do you know if it is possible to restrict ignores to matching errors too?

tests/loo/test_reloo.py
```diff
@@ -0,0 +1,281 @@
...
+
+azb = importorskip("arviz_base")
+xr = importorskip("xarray")
+
+from arviz_stats import loo, reloo
+from arviz_stats.loo.wrapper import SamplingWrapper
+from arviz_stats.utils import ELPDData
+
+
+@pytest.mark.filterwarnings("ignore:Estimated shape parameter:UserWarning")
```
OriolAbril: this looks like the pattern to ignore warnings with matching message, TIL ❤️

OriolAbril (approved):
I think we can merge already as it is a clear improvement on the current situation, but I think we still need to work on that, not so much in terms of coverage number or tests themselves but more organizational level things. This was already an issue before, but I am still not sure if I make a change to the kde where should new tests go, or without running them even where tests to update will be.

tests/loo/test_loo_expectations.py
jordandeklerk: The warning messages were getting wild in the test output because we're explicitly trying to raise a lot of these warnings. So I tried to go through and make sure we're not ignoring organic warnings that surface normally.

jordandeklerk:
> I think we can merge already as it is a clear improvement on the current situation, but I think we still need to work on that, not so much in terms of coverage number or tests themselves but more organizational level things. This was already an issue before, but I am still not sure if I make a change to the kde where should new tests go, or without running them even where tests to update will be.

Agreed, I think the organization could definitely be better. Also, the workflows you mentioned in slack for running different tests suites based on dependencies would be very nice. The tests take slightly longer to run now given the new volume of tests, but I don't think it's too crazy.

PR #221, opened 2025-10-16
DOC: Update the "See Also" parts of API Reference

star1327p:
Update the "See Also" parts of API Reference

----
📚 Documentation preview 📚: https://arviz-stats--221.org.readthedocs.build/en/221/

### DIFF
--- a/src/arviz_stats/summary.py
+++ b/src/arviz_stats/summary.py
@@ -71,8 +71,8 @@ def summary(
     rhat : Compute estimate of rank normalized split R-hat for a set of traces.
     ess : Calculate the effective sample size of a set of traces.
     mcse : Calculate Markov Chain Standard Error statistic.
-    arviz_plots.plot_ess : Plot quantile, local or evolution of effective sample sizes (ESS).
-    arviz_plots.plot_mcse : Plot quantile, local or evolution of Markov Chain Standard Error (MCSE).
+    plot_ess : Plot quantile, local or evolution of effective sample sizes (ESS).
+    plot_mcse : Plot quantile, local or evolution of Markov Chain Standard Error (MCSE).
 
 
     Examples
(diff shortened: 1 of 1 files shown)


src/arviz_stats/summary.py
```diff
@@ -68,11 +68,11 @@ def summary(
...
     See Also
     --------
-    arviz.rhat : Compute estimate of rank normalized split R-hat for a set of traces.
-    arviz.ess : Calculate the effective sample size of a set of traces.
-    arviz.mcse : Calculate Markov Chain Standard Error statistic.
-    plot_ess : Plot quantile, local or evolution of effective sample sizes (ESS).
-    plot_mcse : Plot quantile, local or evolution of Markov Chain Standard Error (MCSE).
+    arviz_stats.rhat : Compute estimate of rank normalized split R-hat for a set of traces.
+    arviz_stats.ess : Calculate the effective sample size of a set of traces.
+    arviz_stats.mcse : Calculate Markov Chain Standard Error statistic.
```
jordandeklerk: We could probably drop the arviz_stats here. I don't see anywhere else where we do that. This should still link to the relevant functions in the API. 

```suggestion
    rhat : Compute estimate of rank normalized split R-hat for a set of traces.
    ess : Calculate the effective sample size of a set of traces.
    mcse : Calculate Markov Chain Standard Error statistic.
```

src/arviz_stats/summary.py
```diff
@@ -68,11 +68,11 @@ def summary(
...
+    arviz_plots.plot_ess : Plot quantile, local or evolution of effective sample sizes (ESS).
+    arviz_plots.plot_mcse : Plot quantile, local or evolution of Markov Chain Standard Error (MCSE).
```
jordandeklerk: This doesn't actually link to anything, but this is the same thing we do in `loo.py` for `plot_compare()` and also in `compare.py`, for example. Not exactly sure if it should be left like that. Probably fine for now though.

jordandeklerk (approved):

PR #226, opened 2025-10-21
Fix stacking weights computation

aloctavodia:
* Incorporates the fix contributed in https://github.com/arviz-devs/arviz/pull/2492
* Avoid updating the iterated dictionary.

----
📚 Documentation preview 📚: https://arviz-stats--226.org.readthedocs.build/en/226/


jordandeklerk:
I haven't looked too deeply, but I wonder if this is an issue on the R side in the `loo` package as well.

Issue #140, opened 2025-06-11
Improve CI configuration


OriolAbril:
We have multiple runs/workers for different python versions, with all of them running the full test suite and one of them also running the `tox -e check` command and the minimal tests.

As we integrate more of the things we have in arviz-base (for now) we'll need to improve that setting I think. Also for things that are specific to arviz-stats. Some examples:

* `arviz-base` also runs tests with [nightlies of scientific-python libraries](https://anaconda.org/scientific-python-nightly-wheels)
* `arviz-base` already uses `docstub` and should have static type checker runs as part of its CI soon-ish https://github.com/arviz-devs/arviz-base/issues/67
* `arviz-stats` is the most bleeding edge out of the 3 libraries when it comes to numpy+scipy usage. I think it would be very nice to run tests on the oldest supported versions to make sure they are actually supported and we aren't using a new numpy/scipy feature without realizing it.

Issue #33, opened 2024-10-18
Add tests and improve CI infrastructure


OriolAbril:
We should have multiple enviornments on which to run tests, with and without optional dependencies, as well as testing with scientific python nightlies in one of these envs.

PR #230, opened 2025-10-27
extend CI jobs and tox settings

OriolAbril:
related to #140 and #33

----
📚 Documentation preview 📚: https://arviz-stats--230.org.readthedocs.build/en/230/

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -83,7 +83,10 @@ setenv =
   NIGHTLIES=TRUE
 
 [testenv:nightlies-xarray]
-base = nightlies
+base =
+    nightlies
+    xarray
+    testenv
 
 [testenv:check]
 description = perform style checks


tox.ini
```diff
@@ -2,45 +2,70 @@
 envlist =
     check
     docs
-    minimal
-    {py311,py312,py313}{,-coverage}
+    {minimal,nonumba,alltests}{,-coverage}
```
OriolAbril: do not hesitate to suggest better names for these. This will also be a "workflow breaking change" and have `arviz-stats` differ from the other two repos, but I didn't know how to conciliate this extended configuration with the `py3XY` job names. If we like `alltests` or if we prefer `test` for the recommended full-fledged test suite we can adapt the other two to follow a similar pattern (`tox -e alltests` instead of `py3XY` depending on the python you are in)

OriolAbril:
I think this is ready to merge now 🚀

PR #233, opened 2025-10-29
Fix return docstrings for loo functions

jordandeklerk:
Standardizes the return docstrings for various LOO functions

----
📚 Documentation preview 📚: https://arviz-stats--233.org.readthedocs.build/en/233/


OriolAbril (approved):
Looks good, thanks. I think we should also add the `ELPDData` class to the docs but we can do it somewhere else

Issue #232, opened 2025-10-27
Unfinished References Section in `arviz_stats.rhat`


star1327p:
Unfinished References Section in `arviz_stats.rhat`:
https://arviz-stats.readthedocs.io/en/latest/api/generated/arviz_stats.rhat.html

We need to include the full citation for each reference paper.

<img width="491" height="158" alt="Image" src="https://github.com/user-attachments/assets/3770f747-1203-458e-9916-d55a82a8b2cd" />

aloctavodia:
I thinks is OK to just include Vehtari et al. (2021). Rank-normalization, folding, and localization: An improved Rhat for assessing convergence of MCMC. Bayesian analysis, 16(2):667-718. we could add a "and references therein"...

PR #236, opened 2025-10-30
Remove extra references in `arviz_stats.rhat`

star1327p:
Remove extra references in `arviz_stats.rhat`:
https://arviz-stats.readthedocs.io/en/latest/api/generated/arviz_stats.rhat.html

Fixes Issue #232.

### DIFF
--- a/src/arviz_stats/sampling_diagnostics.py
+++ b/src/arviz_stats/sampling_diagnostics.py
@@ -228,15 +228,13 @@ def rhat(
     greater than one indicate that one or more chains have not yet converged.
 
     Rank values are calculated over all the chains with ``scipy.stats.rankdata``.
-    Each chain is split in two and normalized with the z-transform following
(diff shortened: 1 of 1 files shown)


src/arviz_stats/sampling_diagnostics.py
```diff
@@ -236,9 +236,6 @@ def rhat(
     * Vehtari et al. (2021).  Rank-normalization, folding, and
       localization: An improved Rhat for assessing convergence of
       MCMC. Bayesian analysis, 16(2):667-718.
```
jordandeklerk: I wonder if we should do something like this to be consistent with the other citations in the rest of the arviz-stats functions:

```suggestion
    .. [1] Vehtari et al. *Rank-normalization, folding, and localization: An improved Rhat 
        for assessing convergence of MCMC*. Bayesian Analysis, 16, 2 (2021). 
        https://doi.org/10.1214/20-BA1221
```

Issue #234, opened 2025-10-30
Add order statistic check for `compare()`


jordandeklerk:
We should implement an order statistics-based diagnostic check in `compare()` like this https://github.com/stan-dev/loo/blob/d6fe380161fcd3ba07065ce0a525146abdb2c1d7/R/loo_compare.R#L300 to warn users when model differences may be due to chance rather than genuine predictive performance differences.

From this paper https://arxiv.org/pdf/2309.03742, when comparing many models using cross-validation, the "best" model can appear better simply due to random variation in ELPD estimates. The more models compared, the higher the probability that one will randomly score highest.

PR #237, opened 2025-10-30
Add order statistic diagnostic for `compare()`

jordandeklerk:
This implements the order statistic diagnostic for detecting selection induced bias when comparing many models in `compare()`. When more than 11 models are compared using full PSIS-LOO-CV, the diagnostic estimates whether the observed performance difference could be due to chance by comparing the best model's ELPD difference against the expected maximum order statistic under a null hypothesis of equal performance. 

This diagnostic is not done for the subsampling case. As far as I can tell, it isn't done here either https://github.com/stan-dev/loo/blob/d6fe380161fcd3ba07065ce0a525146abdb2c1d7/R/loo_compare.psis_loo_ss_list.R. I'm not exactly sure why this is, but my guess is that the the theoretical assumptions underlying the test don't account for the additional variance and approximation bias introduced by subsampling, making it unclear whether the null distribution would be properly calibrated in that setting.

---
Resolves https://github.com/arviz-devs/arviz-stats/issues/234

### DIFF
--- a/src/arviz_stats/loo/compare.py
+++ b/src/arviz_stats/loo/compare.py
@@ -558,14 +558,12 @@ def _order_stat_check(ics_dict, model_order, has_subsampling):
     baseline_model = model_order[baseline_idx]
     baseline_elpd = ics_dict[baseline_model]
 
-    elpd_diffs = []
-    for model_name in model_order:
-        if model_name == baseline_model:
-            elpd_diffs.append(0.0)
-        else:
-            elpd_a_vals = np.asarray(baseline_elpd.elpd_i).flatten()
-            elpd_b_vals = np.asarray(ics_dict[model_name].elpd_i).flatten()
-            elpd_diffs.append(np.sum(elpd_b_vals - elpd_a_vals))
+    elpd_diffs = np.zeros(len(model_order))
+    for idx, model_name in enumerate(model_order):
+        if model_name != baseline_model:
+            elpd_a_vals = np.ravel(baseline_elpd.elpd_i)
+            elpd_b_vals = np.ravel(ics_dict[model_name].elpd_i)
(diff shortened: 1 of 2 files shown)


src/arviz_stats/loo/compare.py
```diff
@@ -23,8 +23,20 @@ def compare(
...
+    is performed to check whether differences in predictive performance are
+    potentially due to chance, as described by [1]_. A warning is issued if
+    selection-induced bias is detected. If this is the case, we recommend avoiding
+    model selection based on LOO-CV and instead favor model averaging/stacking
+    or projection predictive inference.
+
+    See the EABM chapters on `Model Comparison <https://arviz-devs.github.io/EABM/Chapters/Model_comparison.html>`_,
+    `Model Comparison (Case Study) <https://arviz-devs.github.io/EABM/Chapters/Case_study_model_comparison.html>`_,
+    and `Model Comparison for Large Data <https://arviz-devs.github.io/EABM/Chapters/Model_comparison_large_data.html>`_
+    for more details.
```
jordandeklerk: Open to re-wording/removing this however we see fit

tests/loo/test_compare.py
```diff
@@ -411,3 +411,72 @@ def test_compare_elpd_diff_relative_to_best(centered_eight, non_centered_eight):
     for i in range(len(result)):
         expected_diff = best_elpd - result.iloc[i]["elpd"]
         assert_almost_equal(result.iloc[i]["elpd_diff"], expected_diff, decimal=10)
+
+
+def test_compare_order_stat_check(centered_eight, rng):
+    models = {}
+    for i in range(12):
+        loo_result = loo(centered_eight, pointwise=True)
```
aloctavodia: This could be out of the loop

aloctavodia (approved):

Issue #223, opened 2025-10-17
Test numpy version of `logsumexp`


jordandeklerk:
In light of https://github.com/arviz-devs/arviz-stats/issues/219, we should test out https://github.com/arviz-devs/arviz-stats/blob/main/src/arviz_stats/base/stats_utils.py#L164 to see if there are any pain points with this numpy-only version compared to the scipy implementation.

PR #238, opened 2025-10-31
Add tests for numpy logsumexp

jordandeklerk:
Adds more tests for numpy `logsumexp`. This seems to be very well in-line with the scipy version.

---
Resolves https://github.com/arviz-devs/arviz-stats/issues/223

### DIFF
--- a/tests/base/test_stats_utils.py
+++ b/tests/base/test_stats_utils.py
@@ -305,7 +305,7 @@ def test_logsumexp_loo_shapes(rng, shape, axis, expected_shape):
 
     max_log_lik = np.max(log_lik, axis=axis)
     assert np.all(scipy_result <= max_log_lik + 1e-6)
-    assert np.all(scipy_result >= max_log_lik - np.log(n_samples)  - 1e-6)
+    assert np.all(scipy_result >= max_log_lik - np.log(n_samples) - 1e-6)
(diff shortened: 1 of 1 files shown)


jordandeklerk (commented):
Happy to expand and modify these tests if we're not capturing something crucial in the comparison against Scipy.

tests/base/test_stats_utils.py
```diff
@@ -281,3 +281,210 @@ def test_logsumexp_edge_b():
...
+    scipy_result = logsumexp(log_lik, axis=axis, b=1 / n_samples)
+    arviz_result = _logsumexp(log_lik, axis=axis, b=1 / n_samples)
+
+    assert scipy_result.shape == expected_shape
+    assert arviz_result.shape == expected_shape
+    assert_array_almost_equal(scipy_result, arviz_result, decimal=10)
+
+    max_log_lik = np.max(log_lik, axis=axis)
+    assert np.all(scipy_result <= max_log_lik + 1)
+    assert np.all(scipy_result >= -10)
```
aloctavodia:
```suggestion
    assert np.all(scipy_result <= max_log_lik + 1e-6)
    assert np.all(scipy_result >= max_log_lik - np.log(n_samples)  - 1e-6)
```
We may want to use tighter bounds. But also, we are already testing value per value against scipy, so probably not that important.

PR #242, opened 2025-11-07
Fix r2_score and add more R2 flavors

aloctavodia:
The `r2_score` function computed the R² in terms of the predictive distribution, but in the [Bayesian R2 paper](http://www.stat.columbia.edu/~gelman/research/unpublished/bayes_R2_v3.pdf), the definition is in terms of the predicted means (not the predicted observations). See also this [case study](https://avehtari.github.io/bayes_R2/bayes_R2.html). So in this PR I am changing that.

This also renames `r2_score` to `residual_r2`, and adds the `bayesian_r2` and `loo_r2` versions too. 

I am also including circular versions for all these R2, which are not discussed in the paper or the Case Study, so extra attention reviewing this part is welcome ~(I will double-check myself ASAP)~. I decided to use a different definition of R2. Otherwise, for residual and bayesian cases the R2 does not necessarily approach 0 for a model that is pure noise; the reason is that on the circle and for low values of kappa, we can not increase the residual or modeled variance without also increasing the variance of mu, which "inflates" the R2, which may make it difficult to interpret. Use a different formula, but it can also be problematic.

I compared the results to the first example in the case study, and I am getting the same results for `residual_r2` and `bayesian_r2`, ~but different for `loo_r2` (0.56 vs 0.44), so this also needs extra eyes~ (edited: I found a missplaced parenthesis in the code for `loo_r2` in the case study). Notice I am not using the same posterior (instead, I am replicating the model with Bambi).

Depends on https://github.com/arviz-devs/arviz-base/pull/115

### DIFF
--- a/src/arviz_stats/base/diagnostics.py
+++ b/src/arviz_stats/base/diagnostics.py
@@ -8,8 +8,8 @@ import numpy as np
 from scipy import stats
 from scipy.special import logsumexp
 
+from arviz_stats.base.circular_utils import circular_diff, circular_var
 from arviz_stats.base.core import _CoreBase
-from arviz_stats.base.stats_utils import _circdiff, _circular_var
 from arviz_stats.base.stats_utils import not_valid as _not_valid
 
 
@@ -782,6 +782,9 @@ class _DiagnosticsBase(_CoreBase):
 
         """
         if scale is None:
+            if circular:
+                raise ValueError("scale must be provided for circular response.")
+
             # Bernoulli-like models: use Tjur’s pseudo-variance
             scale = np.mean(mu_pred * (1 - mu_pred), axis=1)
         else:
@@ -789,7 +792,7 @@ class _DiagnosticsBase(_CoreBase):
                 raise ValueError("scale_kind must be either 'sd' or 'var'")
 
             if np.any(scale < 0):
-                raise ValueError("Variance must be non-negative.")
+                raise ValueError("Scale must be non-negative.")
 
             if scale_kind == "sd":
                 if circular:
@@ -820,7 +823,7 @@ class _DiagnosticsBase(_CoreBase):
         array-like  (sample, dims)
         """
         if circular:
-            return 1 - _circular_var(_circdiff(y_obs, mu_pred))
+            return 1 - circular_var(circular_diff(y_obs, mu_pred))
 
         var_y_est = np.var(mu_pred, axis=1, ddof=1)
         var_e = np.var(y_obs - mu_pred, axis=1, ddof=1)
--- a/src/arviz_stats/loo/loo_expectations.py
+++ b/src/arviz_stats/loo/loo_expectations.py
@@ -6,7 +6,7 @@ from arviz_base import convert_to_datatree, extract, rcParams
 from scipy.stats import dirichlet
 from xarray import apply_ufunc
 
-from arviz_stats.base.stats_utils import _circdiff, _circular_var
+from arviz_stats.base.circular_utils import circular_diff, circular_mean, circular_sd, circular_var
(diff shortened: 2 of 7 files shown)


src/arviz_stats/base/diagnostics.py
```diff
@@ -756,22 +758,75 @@ def _rhat_nested(ary, superchain_ids):
...
+        Returns
+        -------
+        array-like  (sample, dims)
+        """
+        if circular:
+            var_y_est = circvar(mu_pred, axis=1, high=np.pi, low=-np.pi)
+            var_e = circvar(_circdiff(y_obs, mu_pred), axis=1, high=np.pi, low=-np.pi)
+        else:
+            var_y_est = np.var(mu_pred, axis=1)
+            var_e = np.var(y_obs - mu_pred, axis=1)
```
jordandeklerk: Same variance question as above.

src/arviz_stats/loo/loo_expectations.py
```diff
@@ -230,6 +232,122 @@ def loo_metrics(data, kind="rmse", var_name=None, log_weights=None, round_to="2g
...
+    var_name,
+    n_simulations=4000,
+    summary=True,
+    point_estimate=None,
+    ci_kind=None,
+    ci_prob=None,
+    circular=False,
+    round_to="2g",
+):
+    """Compute LOO-adjusted R².
```
jordandeklerk: Same math in docstring comment

src/arviz_stats/metrics.py
```diff
@@ -11,35 +11,62 @@
...
+    scale_kind="sd",
     summary=True,
     point_estimate=None,
     ci_kind=None,
     ci_prob=None,
+    circular=False,
     round_to="2g",
 ):
-    """R² for Bayesian regression models.
+    r"""Bayesian R² for regression models.
```
jordandeklerk: Same math in docstring comment

src/arviz_stats/metrics.py
```diff
@@ -86,28 +140,134 @@ def r2_score(
...
+    Returns
+    -------
+    Namedtuple or array
 
-        return r2_summary(estimate, c_i[0], c_i[1])
+
+    See Also
+    --------
+    arviz_stats.bayesian_r2 : Bayesian R².
+    arviz_stats.loo_r2 : LOO-adjusted R².
```
jordandeklerk: Same math in docstring comment

src/arviz_stats/metrics.py
```diff
@@ -86,28 +140,134 @@ def r2_score(
...
 
     if summary:
-        estimate = getattr(np, point_estimate)(r_squared).item()
-        c_i = getattr(array_stats, ci_kind)(r_squared, ci_prob)
+        return _summary_r2(r_squared, point_estimate, ci_kind, ci_prob, round_to)
+
+    return r_squared
+
+
+def residual_r2(
```
jordandeklerk: One thought: might be helpful to clarify in the docs how this differs from the paper's recommendation, since it uses empirical residuals rather than modeled variance. From what I can see, we have

`bayesian_r2`:
```python
r2 = var_y_est / (var_y_est + scale)  # scale = modeled variance
```

`residual_r2`:
```python
var_e = np.var(y_obs - mu_pred, axis=1)  # empirical residuals from data
r_squared = var_y_est / (var_y_est + var_e)
```

In the paper they mention, "This measure is model based: all variance terms come from the model, and not directly from the data." So `residual_r2` uses `y_obs - mu_pred` which is observed data instead of the model's expected residual variance. What do you think?

jordandeklerk (approved):
LGTM! Left some comments regarding the circular path. I don't see any tests for the circular implementation for any of the R^2 variants. Might be good to check some things there.

PR #246, opened 2025-11-13
Add original Pareto k to `ELPDData` for moment matching

jordandeklerk:
This adds the original Pareto k pointwise values as well as the Pareto k values after moment matching. It would be useful for the user to be able to compare these after moment matching.


src/arviz_stats/utils.py
```diff
@@ -191,6 +191,7 @@ class ELPDData:  # pylint: disable=too-many-ancestors, too-many-instance-attribu
     loo_subsample_observations: np.ndarray = None
     elpd_loo_approx: DataArray = None
     log_jacobian: DataArray = None
+    pareto_k_original: DataArray = None
```
jordandeklerk: Open to other names for this

jordandeklerk: Switched to that

PR #251, opened 2025-11-19
Add array-only interface for `loo()`

jordandeklerk:
This implements array interfaces for three PSIS-LOO-CV functions, `loo()`, `loo_approximate_posterior()` and `loo_score()`.

Added a new function `process_chain_none_multi` to cleanly handle processing multiple arrays when `chain_axis` is `None`.

### DIFF
--- a/src/arviz_stats/base/array.py
+++ b/src/arviz_stats/base/array.py
@@ -788,7 +788,7 @@ class BaseArray(_DensityBase, _DiagnosticsBase):
 
     def loo_approximate_posterior(
         self,
-        log_likelihood,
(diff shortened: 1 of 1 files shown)


aloctavodia:
This is good. For many functions in arviz-stats, the computations are implemented once at the array level, and the higher-level functions then call the array interface. This avoids code duplication. The way we implement this varies, as some functions require access to several groups, and others operate on single groups.  I wonder if we want to replicate that approach here or if we have reasons to avoid it. For instance, now `_compute_loo_results` has code duplicated with what is implemented here.

PR #256, opened 2025-11-21
Use faster cdf for hypergeometric

aloctavodia:


aloctavodia:
`plot_rank` and `plot_rank_dist` are 2 to 3 times faster.

Issue #180, opened 2025-08-22
Add support for Mixture IS LOO


jordandeklerk:
See the original issue here https://github.com/stan-dev/loo/issues/270. We can do something very similar to the approach outlined in the original issue. 

This is yet another useful tool we could add to the ArviZ ecosystem to generate more accurate LOO-CV and ELPD estimates when we have many large Pareto-k values.

PR #260, opened 2025-11-26
Add mixture importance sampling LOO-CV support

jordandeklerk:
Adds a `mixture` parameter to the loo function for computing LOO-CV when samples come from a mixture posterior distribution. See https://mc-stan.org/loo/articles/loo2-mixis.html for reference. Based on [Silva and Zanella (2022)](https://arxiv.org/abs/2209.09190).

We should add an example for using this in EABM and then update the warning with a link to the example once it's completed.

---
Resolves https://github.com/arviz-devs/arviz-stats/issues/180

### DIFF
--- a/tests/loo/test_loo.py
+++ b/tests/loo/test_loo.py
@@ -1,7 +1,6 @@
 """Test PSIS-LOO-CV."""
 
 # pylint: disable=redefined-outer-name, unused-argument
-import numpy as np
 import pytest
 from numpy.testing import assert_allclose, assert_almost_equal, assert_array_equal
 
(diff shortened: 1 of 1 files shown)


src/arviz_stats/loo/loo.py
```diff
@@ -161,6 +175,63 @@ def loo(
...
 
+    if mixture:
+        warnings.warn(
+            "Using mixture=True assumes the log-likelihood values were computed from samples "
+            "drawn from a mixture posterior distribution. "
+            "See Silva and Zanella (2023) https://doi.org/10.1080/01621459.2023.2257893 "
+            "for details on how to sample from the mixture posterior.",
+            UserWarning,
+            stacklevel=2,
+        )
```
jordandeklerk: Open to suggestions about how to word this

aloctavodia: Sounds clear to me. Is there something in particular that makes you doubt? Once we have an example in EABM, we should like to the example. And that will make things clear.

PR #267, opened 2025-12-04
Centralize PSIS-LOO-CV computations in `array` layer

jordandeklerk:
This is the next phase of https://github.com/arviz-devs/arviz-stats/pull/251.

Refactors core PSIS-LOO computations to follow a single computational layer where all `numpy` computations live in `diagnostics.py`, with `array.py` providing `ufunc` wrappers and `dataarray.py`/`accessors.py` providing the `xarray` interface. Trying to avoid any duplicate computations throughout the PSIS-LOO-CV universe.

Also adds Mix-IS-LOO (`loo_mixture`) support through the full stack, and fixes a circular import between `helper_loo.py` and `loo_approximate_posterior.py`. 

Moving forward, I think `helper_loo.py` will continue to serve purely as an orchestration layer responsible for data extraction, validation, and result packaging while delegating all statistical computations to accessors that call through the array layer. Any pure `numpy` computations will continue to be migrated to `diagnostics.py` and exposed via new accessor methods in future PRs.

### DIFF
--- a/src/arviz_stats/base/array.py
+++ b/src/arviz_stats/base/array.py
@@ -936,7 +936,7 @@ class BaseArray(_DensityBase, _DiagnosticsBase):
         mix_log_weights : array-like
             Mixture log weights
         """
-        ary = np.asarray(ary)
+        ary, chain_axis, draw_axis = process_chain_none(ary, chain_axis, draw_axis)
         ndim = ary.ndim
         chain_axis = chain_axis % ndim
         draw_axis = draw_axis % ndim


src/arviz_stats/base/dataarray.py
```diff
@@ -395,6 +395,244 @@ def pareto_khat(self, da, sample_dims=None, r_eff=None, tail="both", log_weights
...
+            da,
+            log_p,
+            log_q,
+            input_core_dims=[dims, dims, dims],
+            output_core_dims=[[], [], []],
+            vectorize=True,
+            kwargs=kwargs,
+        )
+
+    def loo_mixture(self, da, sample_dims=None, log_jacobian=None):
```
jordandeklerk: Note that this departs slightly from the standard use of `apply_ufunc` here like the other functions. The Mix-IS-LOO computation has a cross-observation dependency since the importance weight for observation `i` depends on the log-likelihoods of all other observations. So the `logsumexp` across `obs_axes` means this needs the full array at once, e.g., we can't vectorize these computations per-observation.

Maybe I'm missing something so open to thoughts on this though.

PR #268, opened 2025-12-06
Add numba to docs tox environment

jordandeklerk:
The local docs build with `tox -e docs` fails because we don't have the `numba` dependency. If we don't need this, I think another solution would be to use `autodoc_mock_imports` in the sphinx config to mock the `numba` dependency. Need to double check this though.

If this isn't necessary we can close this, but I personally like to build the docs locally to catch anything before opening a PR.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -108,6 +108,7 @@ setenv =
 extras =
     doc
     xarray
+    numba
 commands =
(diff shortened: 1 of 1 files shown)


OriolAbril (approved):
I prefer adding numba instead of mocking the import too. It used to be pulled in by the `xarray` bundle of optional dependencies but that is not the case anymore. Thanks

PR #286, opened 2026-01-16
Fix CI after arviz-base and h5netcdf updates

tomicapretto:


pyproject.toml
```diff
@@ -41,7 +41,7 @@ funding = "https://opencollective.com/arviz"
 [project.optional-dependencies]
 # user facing optional dependency blocks
 xarray = [
-  "arviz-base>=0.8.0,<0.9.0",
+  "arviz-base>=0.8.2,<0.9.0",
```
OriolAbril: I think we can skip this given I don't think it is worth doing a 0.8.1 release for arviz-stats (which means this will soon become a dependency on github install), but it is perfectly fine either way

PR #304, opened 2026-02-25
Add custom rounding rules to compare

### LINKED ISSUE #300: Show less digits for compare
We should have some rounding rules for `compare` similar to those we already have for summary. @sethaxen do you have any suggestions?

aloctavodia:
Closes #300 

With the new default, `round_to="auto"`, we get 

```
	      rank	elpd	  p	  elpd_diff	weight	se	     dse	warning
hurdle_nb	0	-370.0	6.2	      0.0	0.940	14.0	 0.0	False
poisson	    1	-470.0	16.5	100.0	0.062	25.0	20.0	False
```

instead of round_to=None (equivalent to previous default)

```

         rank	      elpd	     p	    elpd_diff	weight	    se	         dse	   warning
hurdle_nb	0	-372.763715	6.241041	0.000000	0.938125	13.698339	0.000000	False
poisson	    1	-468.657580	16.471540	95.893866	0.061875	24.842769	19.514716	False

```

### DIFF
--- a/tests/loo/test_compare.py
+++ b/tests/loo/test_compare.py
@@ -233,7 +233,7 @@ def test_compare_single_model(centered_eight, method):
 @pytest.mark.parametrize("method", ["stacking", "BB-pseudo-BMA", "pseudo-BMA"])
 def test_compare_many_models(centered_eight, method):
     models = {f"model_{i}": centered_eight for i in range(7)}
-    result = compare(models, method=method, round_to=None)
+    result = compare(models, method=method, round_to="None")
 
     assert len(result) == 7
     assert result["rank"].min() == 0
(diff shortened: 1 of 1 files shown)


src/arviz_stats/loo/compare.py
```diff
@@ -150,6 +162,11 @@ def compare(
         Bayesian Analysis, 13, 3 (2018). https://doi.org/10.1214/17-BA1091
         arXiv preprint https://arxiv.org/abs/1704.02030.
     """
+    if round_to == "auto":
```
OriolAbril: This seems to be mixing auto as in "special rounding rules" with auto as in use whatever rcParams says.

Also, not sure if this ended up outdated but https://python.arviz.org/projects/base/en/latest/api/index.html#stats.round_to says the rcParam is 2g which would mean we can not trigger the if auto later on unless the rcParam is modified.

aloctavodia: I agree we need to make it more clear for summary and compare. For these functions we have a set of customs rules that uses the value from rcParams["stats.round_to"] plus some other specifications, for other functions like mean we directly usercParams["stats.round_to"].

Issue #177, opened 2025-08-19
Add PSIS-LFO-CV


OriolAbril:
Original issue: https://github.com/arviz-devs/arviz/issues/2068

jordandeklerk:
I can work on this if you want

PR #305, opened 2026-02-25
Add PSIS LFO-CV

TeemuSailynoja:
This is a work in progress PR for adding approximate leave-future-out cross validation (LFO-CV) using pareto-smoothed importance sampling (PSIS).

The current implementation was planned by me and mostly implemented by Agentic assistance. I still need to personally go through it in detail.

- [ ] Check computations in 9460737ccbb509be9f2a471332f361386089a9a7 
- [ ] Ensure all tests in 6dc74cbf90fa8046fdf0f859ccc44d602a53f55e are needed and no gaps left
- [ ] Make another branch for an example notebook to run a full case study to test that everything works as intended.

This PR is related to issue #177


TeemuSailynoja:
@jordandeklerk I initiated a PR. I'm also very busy, but wanted to get past the "empty page"  blocker and have a start for the PR :)

jordandeklerk:
> @jordandeklerk I initiated a PR. I'm also very busy, but wanted to get past the "empty page" blocker and have a start for the PR :)

Awesome, I will take a look soon!

jordandeklerk:
Reminder to myself to revisit this. Have some more time now in the coming months. Looking forward to getting this into the ArviZ ecosystem!

jordandeklerk:
@TeemuSailynoja Made some changes since you first opened this. I think we can open this for review now and get some other eyes on this. Would love to get this into the ArviZ ecosystem ASAP.

Quick summary of what changed:

* Split the implementation into `lfo_cv.py` and `lfo_cv_helper.py` to follow the existing PSIS-LOO-CV structure.
* Fixed the sign of the PSIS importance ratios. `psislw` negates its input internally for the LOO case, so LFO needs the summed log likelihoods passed with the opposite sign. This was causing incorrect approximate elpds and too many refits.
* Refit origins are now scored exactly, `refits` only records PSIS-triggered refits, and `r_eff` is estimated from each refit posterior rather than implicitly using 1.
* Added `compare` support, including checks for incompatible LFO settings and comparisons against LOO or kfold. I also fixed a pre-existing `_calculate_ics` crash when `pareto_k` is `None`.

## M-SAP scoring

Something interesting happens when `M > 1`, which I think is the main thing worth discussing.

For a forecast horizon of `M`, `lfo_cv` scores the block using the chain rule

$$
p(y_{i+1:i+M} \mid y_{1:i}) = \prod_{m=1}^{M} p(y_{i+m} \mid y_{1:i+m-1}).
$$

So once we move past the first observation in the block, the next prediction conditions on the observations that came before it.

This is slightly different from what happens in the `loo` Lake Huron vignette [here](https://github.com/stan-dev/loo/blob/a59519c224d592594f00489e7b705b799699e1a5/vignettes/loo2-lfo.Rmd#L515-L527) and [here](https://github.com/stan-dev/loo/blob/a59519c224d592594f00489e7b705b799699e1a5/vignettes/loo2-lfo.Rmd#L540-L582). As far as I can see, the vignette passes the full `M`-step block to `log_lik(..., oos = oos)` and then sums those log likelihoods. For an autoregressive model, though, `oos` has some important behavior inside brms that I think is easy to miss.

In [`prepare_predictions_ac`](https://github.com/paul-buerkner/brms/blob/29ccd2384776b2bfc13bf749659528911743a92b/R/prepare_predictions.R#L742-L750), brms sets the responses in `oos` to `NA` for the AR predictor. Then in [`.predictor_arma`](https://github.com/paul-buerkner/brms/blob/29ccd2384776b2bfc13bf749659528911743a92b/R/predictor.R#L514-L525), those missing responses are generated from the posterior predictive distribution and their residuals are carried forward into the later AR predictions. So for `M > 1`, later points in the block are predicted using simulated earlier responses rather than the observed earlier responses.

It seems to me that these are two reasonable but different prediction tasks. The brms approach is closer to generating an `M`-step forecast trajectory from the forecast origin and scoring the observed points along the way. What we currently do is score the joint predictive density of the observed block through its chain rule factorization.

I think the chain rule version is probably the better definition for `lfo_cv` for a few reasons, but happy to hear thoughts on this.

1. It matches the estimand in the paper as far as I can tell. Bürkner, Gabry and Vehtari define the `M`-SAP elpd in terms of the joint predictive density $p(y_{i+1:i+M} \mid y_{1:i})$, and the expression above is exactly its chain rule factorization evaluated at the observed values. The simulated-lag version answers a different question, where the intermediate observations remain unobserved and are simulated instead. It seems to me that this follows from how brms implements `oos` through its forecasting machinery rather than from the definition of the `M`-SAP estimand itself.
2. The chain rule score is deterministic given the posterior draws and observed data. The trajectory version introduces another layer of Monte Carlo variation because each draw carries a simulated path forward. I think avoiding that extra variability is preferable for something that eventually feeds into `compare`.
3. It also fits the generic wrapper interface much better. `log_likelihood__i` asks the wrapper to evaluate the model's log density at observed data. Reproducing the brms behavior would require posterior predictive simulation inside the scoring step, which would expand the wrapper contract quite a bit and would be harder to support consistently across PPLs.
4. I think it is also more consistent with the other cross-validation methods in `arviz-stats`. We are evaluating the density of the data that actually occurred under the relevant predictive distribution. The chain rule block score gives us that quantity for the time-series case, while the trajectory version conditions later predictions on simulated intermediate outcomes.

I wanted to flag this specifically because it changes the quantity being compared for autoregressive models when `M > 1`. For `M = 1`, or for models where the likelihood does not depend on lagged responses, the distinction disappears. If we eventually want to support the forecast-trajectory interpretation too, I think an optional wrapper hook could make sense in a follow-up rather than changing what `forecast_horizon` means here.

---
Also worth noting that it looks like [pymc-forecast](https://github.com/pymc-labs/pymc-forecast) is starting to cover the forecast-trajectory side of this in the PyMC ecosystem. Its `backtest` works over expanding or rolling windows, generates multi-step forecasts through posterior predictive simulation, and scores those forecasts with sample-based metrics such as CRPS. That is much closer to the trajectory semantics used by brms `oos`.

I think that is a useful distinction. If the goal is to evaluate a full forecast trajectory where uncertainty propagates forward through simulated intermediate values, posterior predictive samples and scoring rules like CRPS are a natural fit. `lfo_cv` is just doing something slightly different.

jordandeklerk:
Wanted to also flag that I have https://github.com/arviz-devs/EABM/pull/223 as a draft if you want a walkthrough of the `lfo_cv` functionality as it exists on this branch. This is the chapter that recreates (to the extent that we can) the `loo` vignette https://mc-stan.org/loo/articles/loo2-lfo.html.

Keep in mind that the chapter is still WIP, but the code examples for the most part recreate everything in the `loo` vignette as far as I can tell.

src/arviz_stats/loo/lfo_cv_helper.py
```diff
@@ -0,0 +1,304 @@
...
+    sample_dims = lfo_inputs.sample_dims
+    n_samples = lfo_inputs.n_samples
+    time_dim = lfo_inputs.time_dim
+    horizon = lfo_inputs.forecast_horizon
+
+    origins = np.arange(lfo_inputs.min_observations, lfo_inputs.n_time_points - horizon + 1)
+    elpds = np.empty(len(origins))
+    lpds = np.empty(len(origins))
+    for pos, cutoff in enumerate(origins):
+        log_lik, dims, n_refit, _ = _refit_loglik(lfo_inputs, wrapper, cutoff)
```
aloctavodia: I think the name `n_refit` is potentially confusing. Here is the number of samples in a refit, and elsewhere we have a variable n_refits` that is the number of refits. Is that correct?

aloctavodia (commented):
We also need to update https://github.com/arviz-devs/arviz-stats/blob/cfcd8dab34061c4fada1b86f9db0a4f384034d09/src/arviz_stats/utils.py#L169 right?

TeemuSailynoja:
@jordandeklerk 
> @TeemuSailynoja Made some changes since you first opened this. I think we can open this for review now and get some other eyes on this. Would love to get this into the ArviZ ecosystem ASAP.

"Some" carries a lot of weight there 😅  Amazing work!
I'm going through it at the moment, but I wanted to stop and express how much I appreciate this. Thank you!
 
> * Split the implementation into `lfo_cv.py` and `lfo_cv_helper.py` to follow the existing PSIS-LOO-CV structure.

This was a reason why I was also quiet for some time. I realized that the original draft had missed the point of the existing structure. When I started reviewing it properly, I also noticed that and realized that it needed an overhaul.

> ## M-SAP scoring
> ...
> It seems to me that these are two reasonable but different prediction tasks. The brms approach is closer to generating an `M`-step forecast trajectory from the forecast origin and scoring the observed points along the way. What we currently do is score the joint predictive density of the observed block through its chain rule factorization.
> 
> I think the chain rule version is probably the better definition for `lfo_cv` for a few reasons, but happy to hear thoughts on this.

I agree with the reasons.

> I wanted to flag this specifically because it changes the quantity being compared for autoregressive models when `M > 1`. For `M = 1`, or for models where the likelihood does not depend on lagged responses, the distinction disappears. If we eventually want to support the forecast-trajectory interpretation too, I think an optional wrapper hook could make sense in a follow-up rather than changing what `forecast_horizon` means here.

This wrapper would be a nice one, I think I would find many places to use that, but indeed it sounds more like an incremental PR rather than something that should be crammed into this one.
 

> I think that is a useful distinction. If the goal is to evaluate a full forecast trajectory where uncertainty propagates forward through simulated intermediate values, posterior predictive samples and scoring rules like CRPS are a natural fit. `lfo_cv` is just doing something slightly different.

A wrapper to push LFO predictive distributions through these scoring functions would be really nice to have.

src/arviz_stats/loo/lfo_cv.py
```diff
@@ -0,0 +1,198 @@
...
+    elpd_data = ELPDData(
+        kind="lfo_cv",
+        elpd=lfo_results.elpd,
+        se=lfo_results.se,
+        p=lfo_results.p,
+        n_samples=lfo_inputs.n_samples,
+        n_data_points=lfo_results.n_data_points,
+        scale="log",
+        warning=warning,
+        good_k=good_k,
```
TeemuSailynoja, line 186: Should this just have `good_k = None if method == "exact" else k_threshold`.
Not a hard request. Now good_k appears twice above, but it's just passed forward in the results anyway.

TeemuSailynoja: Also, perhaps @aloctavodia has an opinion, but should we verify that `k_threshold` is a positive float? I don't think there is anything else that a user could put there by accident though.

TeemuSailynoja:
@aloctavodia 
> We also need to update
> 
> https://github.com/arviz-devs/arviz-stats/blob/cfcd8dab34061c4fada1b86f9db0a4f384034d09/src/arviz_stats/utils.py#L169
> 
> right?

Do you think that now that there is another type of ELPD data that has some shared and some unique attributes, it would make sense to subclasses into `ELPDDataLOO`, `ELPDDataKFold`, `ELPDDataSubsample`, and `ELPDDataLFO`?

This could be a refactor following this PR, so that this one just adds the new attributes and printing method to `ELPDData`, and then we continue to another PR that splits the class to subclasses. This would clean up auto complete a bit, as currently, before this PR, 11 / 24 of the attributes are shared between all sources producing ELPDData – and this PR would add five more unique ones.

src/arviz_stats/loo/lfo_cv.py
```diff
@@ -0,0 +1,198 @@
...
+        elpd_i=lfo_results.elpd_i if pointwise else None,
+        pareto_k=lfo_results.pareto_k if (pointwise and method == "approx") else None,
+    )
+
+    elpd_data.forecast_horizon = forecast_horizon
+    elpd_data.min_observations = min_observations
+    elpd_data.refits = np.asarray(lfo_results.refits)
+    elpd_data.n_refits = n_refits
```
TeemuSailynoja: @aloctavodia , so we should update ELPDData to include these attributes?

src/arviz_stats/loo/lfo_cv.py
```diff
@@ -0,0 +1,198 @@
...
+
+    if pointwise:
+        elpd_data.p_lfo_i = lfo_results.p_lfo_i
```
TeemuSailynoja: ... and this 😉

src/arviz_stats/loo/lfo_cv.py
```diff
@@ -0,0 +1,198 @@
...
+        )
+
+    lfo_inputs = _prepare_lfo_inputs(
+        data, var_name, wrapper, min_observations, forecast_horizon, time_dim
+    )
+
+    if method == "exact":
+        lfo_results = _compute_lfo_exact(lfo_inputs, wrapper)
+    else:
+        lfo_results = _compute_lfo_approx(lfo_inputs, wrapper, k_threshold)
```
TeemuSailynoja, line 162: these could already return instances of ELPDData and the logic below could be in the relevant helper function.

src/arviz_stats/loo/lfo_cv_helper.py
```diff
@@ -0,0 +1,304 @@
...
+        "n_time_points",
+        "min_observations",
+        "forecast_horizon",
+    ],
+)
+
+LFOResults = namedtuple(
+    "LFOResults",
+    ["elpd", "se", "p", "n_data_points", "elpd_i", "p_lfo_i", "refits", "pareto_k"],
+)
```
TeemuSailynoja, line 38: if we make `_assemble_results`, and consequently `_compute_lfo_exact` and `_compute_lfo_approx`, return instances of `ELPDData` this class would not be needed.

src/arviz_stats/loo/lfo_cv_helper.py
```diff
@@ -0,0 +1,304 @@
...
+    n_samples : int
+        Number of posterior draws in the refit.
+    idata : DataTree
+        Inference data of the refit returned by the wrapper.
+    """
+    exclude_idx = np.arange(cutoff, lfo_inputs.n_time_points)
+    train_data, excluded_data = wrapper.sel_observations(exclude_idx)
+    idata = wrapper.get_inference_data(wrapper.sample(train_data))
+
+    log_lik = wrapper.log_likelihood__i(excluded_data, idata)
```
TeemuSailynoja: would it make sense here to only evaluate log-likelihood for `cutoff:cutoff + horizon`?

src/arviz_stats/loo/lfo_cv_helper.py
TeemuSailynoja: we could call it `n_refit_samples`, as `n_samples` is already used for the number of posterior draws in the original `idata`.

src/arviz_stats/loo/lfo_cv_helper.py
TeemuSailynoja, line 282: If some one else arrives here after reading just the "exact" path, this is evaluated for `i >= cutoff + horizon`, so that the "approx" path can use it for computing `log_ratios`.

src/arviz_stats/loo/lfo_cv_helper.py
```diff
@@ -0,0 +1,304 @@
...
+        - pareto_k: Per-origin Pareto k values, NaN at the first origin and wherever a
+          refit occurred
+    """
+    ll_full = lfo_inputs.log_likelihood
+    sample_dims = lfo_inputs.sample_dims
+    n_samples = lfo_inputs.n_samples
+    time_dim = lfo_inputs.time_dim
+    horizon = lfo_inputs.forecast_horizon
+
+    origins = np.arange(lfo_inputs.min_observations, lfo_inputs.n_time_points - horizon + 1)
```
TeemuSailynoja: minor: this could be an output of `_prepare_lfo_inputs`.

src/arviz_stats/loo/lfo_cv_helper.py
```diff
@@ -0,0 +1,304 @@
...
+    for pos, cutoff in enumerate(origins):
+        offset = cutoff - last_refit
+        if offset == 0:
+            log_weights, pareto_k = None, np.nan
+        else:
+            log_ratios = ll_star.isel({time_dim: slice(0, offset)}).sum(time_dim)
+            try:
+                log_weights, pareto_k = (-log_ratios).azstats.psislw(dim=star_dims, r_eff=r_eff)
+            except ValueError:
+                log_weights, pareto_k = None, np.inf
```
TeemuSailynoja: Are we catching infinite weights here?

src/arviz_stats/loo/lfo_cv_helper.py
```diff
@@ -0,0 +1,304 @@
...
+            f"Time dimension '{time_dim}' not found in log_likelihood. "
+            f"Available dimensions: {list(log_likelihood.dims)}"
+        )
+
+    obs_dims = [dim for dim in log_likelihood.dims if dim not in sample_dims]
+    if obs_dims != [time_dim]:
+        raise ValueError(
+            f"lfo_cv currently supports a single time dimension. Found observation "
+            f"dimensions {obs_dims}, expected only '{time_dim}'."
+        )
```
TeemuSailynoja, line 71: is this actually a hard no? I haven't done this for multidimensional timeseries, but with PSIS-LOO-CV, multidimensional is possible, but prone to producing high pareto k values when dimensionality is high.

If the answer is "maybe possible, but requires further testing", let's consider that a potential extension.

TeemuSailynoja:
@jordandeklerk I'm happy to also make some of the suggested edits as it seems to all be cosmetic. The calculations all seem correct.

Or, if someone decides they are just needlessly pedantic, I'm also not feeling too bad to skip them. In the end, I just wanted to help @aloctavodia or @OriolAbril if they come to review next.

src/arviz_stats/loo/lfo_cv_helper.py
jordandeklerk: > I think the name `n_refit` is potentially confusing. Here is the number of samples in a refit, and elsewhere we have a variable n_refits` that is the number of refits. Is that correct?

Yep, that’s right. `n_refit` was the number of posterior samples in that particular refit, while `n_refits` is the number of refits. I renamed it to `n_refit_samples`, as Teemu suggested.

jordandeklerk:
> @jordandeklerk
> 
> > @TeemuSailynoja Made some changes since you first opened this. I think we can open this for review now and get some other eyes on this. Would love to get this into the ArviZ ecosystem ASAP.
> 
> "Some" carries a lot of weight there 😅 Amazing work! I'm going through it at the moment, but I wanted to stop and express how much I appreciate this. Thank you!
> 
> > * Split the implementation into `lfo_cv.py` and `lfo_cv_helper.py` to follow the existing PSIS-LOO-CV structure.
> 
> This was a reason why I was also quiet for some time. I realized that the original draft had missed the point of the existing structure. When I started reviewing it properly, I also noticed that and realized that it needed an overhaul.
> 
> > ## M-SAP scoring
> > ...
> > It seems to me that these are two reasonable but different prediction tasks. The brms approach is closer to generating an `M`-step forecast trajectory from the forecast origin and scoring the observed points along the way. What we currently do is score the joint predictive density of the observed block through its chain rule factorization.
> > I think the chain rule version is probably the better definition for `lfo_cv` for a few reasons, but happy to hear thoughts on this.
> 
> I agree with the reasons.
> 
> > I wanted to flag this specifically because it changes the quantity being compared for autoregressive models when `M > 1`. For `M = 1`, or for models where the likelihood does not depend on lagged responses, the distinction disappears. If we eventually want to support the forecast-trajectory interpretation too, I think an optional wrapper hook could make sense in a follow-up rather than changing what `forecast_horizon` means here.
> 
> This wrapper would be a nice one, I think I would find many places to use that, but indeed it sounds more like an incremental PR rather than something that should be crammed into this one.
> 
> > I think that is a useful distinction. If the goal is to evaluate a full forecast trajectory where uncertainty propagates forward through simulated intermediate values, posterior predictive samples and scoring rules like CRPS are a natural fit. `lfo_cv` is just doing something slightly different.
> 
> A wrapper to push LFO predictive distributions through these scoring functions would be really nice to have.

Thanks, glad we’re on the same page! I kept the chain-rule definition for LFO-CV here. I agree that forecast-trajectory scoring would be useful too, but I think that fits better as a separate wrapper in a follow-up. I can open an issue for that as well.

jordandeklerk:
> @aloctavodia
> 
> > We also need to update
> > https://github.com/arviz-devs/arviz-stats/blob/cfcd8dab34061c4fada1b86f9db0a4f384034d09/src/arviz_stats/utils.py#L169
> > 
> > right?
> 
> Do you think that now that there is another type of ELPD data that has some shared and some unique attributes, it would make sense to subclasses into `ELPDDataLOO`, `ELPDDataKFold`, `ELPDDataSubsample`, and `ELPDDataLFO`?
> 
> This could be a refactor following this PR, so that this one just adds the new attributes and printing method to `ELPDData`, and then we continue to another PR that splits the class to subclasses. This would clean up auto complete a bit, as currently, before this PR, 11 / 24 of the attributes are shared between all sources producing ELPDData – and this PR would add five more unique ones.

I like this split. For this PR I’ve added the LFO fields and printing to the existing `ELPDData`. That part is intentionally temporary, then we can handle the method-specific subclasses together in a focused follow-up. @aloctavodia, does that sound good to you?

src/arviz_stats/loo/lfo_cv.py
jordandeklerk: Yeah, changed it. `good_k` is now assigned once based on the selected method. Good catch!

src/arviz_stats/loo/lfo_cv.py
jordandeklerk: I added `forecast_horizon`, `min_observations`, `refits`, and `n_refits` as fields on `ELPDData` and now pass them through the constructor. For now they live on the shared class, then they can move to the LFO-specific subclass in the follow-up.

I can change this though if we don't like this.

src/arviz_stats/loo/lfo_cv.py
jordandeklerk: Added this one too. It’s now an optional `ELPDData` field and is passed through the constructor, with `None` when pointwise output is disabled.

src/arviz_stats/loo/lfo_cv.py
jordandeklerk: I thought about this, but I kept the compute helpers returning `LFOResults` for now. I think it’s useful to keep the numerical calculation separate from the public result construction, warnings, and pointwise handling. Since we’re already planning a result-class follow-up, I’d rather revisit this there than rearrange it twice. 

Does that sound reasonable?

src/arviz_stats/loo/lfo_cv_helper.py
jordandeklerk: Same thought here. I kept LFOResults as the internal computation result for now. Once we have an `ELPDDataLFO` design, we can decide whether this layer is still useful or whether it should go away.

src/arviz_stats/loo/lfo_cv_helper.py
jordandeklerk: Done! `_prepare_lfo_inputs` now computes the forecast origins once, and both the exact and approximate paths reuse them.

src/arviz_stats/loo/lfo_cv_helper.py
jordandeklerk: Great catch! Not reliably before. I added `_psis_lfo_weights` to catch invalid log ratios, `psislw` failures, and invalid PSIS weights or Pareto $k$ values. Those cases now produce `pareto_k=np.inf`, which makes the approximate path refit at that forecast origin. `-inf` log weights are still allowed because they represent zero weight. 

Let me know what you think of this.

src/arviz_stats/loo/lfo_cv_helper.py
jordandeklerk: I don’t think it’s a hard no. For now the implementation expects one joint log-likelihood value per time point, so there’s one joint importance ratio and Pareto $k$ value per forecast origin. I clarified that users can currently combine a multivariate likelihood before calling `lfo_cv`. 

I think native handling of the extra dimensions needs a clearer aggregation contract and tests, so I’d leave that for a follow-up. But I do think this should be a functionality for sure.

src/arviz_stats/loo/lfo_cv_helper.py
jordandeklerk: Good point! I added a comment in the code explaining that the exact path only scores the immediate forecast block, but the approximate path reuses the fit at later origins and therefore needs the likelihood values after cutoff + horizon to compute subsequent importance ratios.

Let me know if it sounds good or if this should be more than a comment.

jordandeklerk:
> @jordandeklerk I'm happy to also make some of the suggested edits as it seems to all be cosmetic. The calculations all seem correct.
> 
> Or, if someone decides they are just needlessly pedantic, I'm also not feeling too bad to skip them. In the end, I just wanted to help @aloctavodia or @OriolAbril if they come to review next.

Thanks for going through all of this, I really appreciate it! I’ve made most of the changes suggested from @TeemuSailynoja and @aloctavodia . I’ve left the `ELPD` subclass idea and multidimensional questions open because I think those are worth settling as part of the follow-up design.

Let me know you all think of this. Happy to go back and change some things.

src/arviz_stats/loo/lfo_cv_helper.py
TeemuSailynoja: looks great!

src/arviz_stats/loo/lfo_cv.py
TeemuSailynoja: I'm happy to leave this for the revisit.

src/arviz_stats/loo/lfo_cv_helper.py
TeemuSailynoja: It's good :)

TeemuSailynoja (commented):
I think this is ready to merge (ping @aloctavodia). Thank you for pushing this @jordandeklerk !

I created follow-up issues for 
- #445
- #444

aloctavodia (approved):

PR #307, opened 2026-02-26
Set histogram density=True by default

aatifjunaid:
This PR sets density=True as the default value for histogram.
Updated default argument in histogram.
Verified consistency across base and numba backends.
All existing tests pass.
Let me know if additional tests are preferred to explicitly check default behavior.


aatifjunaid:
I’ve addressed the requested changes. Please let me know if anything else needs adjustment.

aatifjunaid:
I have made the changes in the tests . Kindly take a look and let me know if I missed anything.

aloctavodia (approved):

Issue #142, opened 2025-06-15
Raise when leftover kwargs are present


OriolAbril:
Most functions in the xarray interface allow kwargs, which are passed to `apply_ufunc` or downstream to control behaviour of other stats functions called as intermediate steps of the one currently being called (e.g. hist/kde in hdi).

For functions that take `**kwargs` and pass them downstream it is generally not a good idea to check kwargs, as the downstream functions might change behaviour from time to time, but functions that consume those kwargs we should make sure we check there are no leftover kwargs.

As of now, there are several places where we take elements from `kwargs` by name and silently ignore extra keys in `kwargs`. For example both `dims="draw"` and `dim="draw"` work without error or warning in `hdi`, but `dims` is silently ignored (you can check in the output that both `chain, draw` have been reduced, the default sample_dims in rcParams)

![Image](https://github.com/user-attachments/assets/4e89a5d0-e8a7-4a78-a8e5-acbd841f533e)

OriolAbril:
#143 addresses this for hdi and kde functions, and adds some minimal tests to check this happens for eti, hdi, kde and histogram functions.

I am not closing the issue because we should check if this happens with other functions but do note that once #143 is merged the snippet above will no longer behave like the screenshot shows. If you are not sure how (or don't want to) open a PR, this issue is a good place to contribute anyway. Make sure you install from GitHub (see [install instructions](https://arviz-stats.readthedocs.io/en/latest/#installation) on the GitHub tab), check which functions raise an error and which ignore extra keyword arguments and comment back here.

PR #314, opened 2026-03-03
Add comprehensive tests for kwargs validation across all functions

Shlokpalrecha:
## What this PR does

Adds test coverage for kwargs validation across arviz-stats functions (relates to #142).

## Background

While working on #142, I tested whether functions properly reject invalid kwargs. Turns out they all work correctly! But I noticed we only have tests for 4 functions (from PR #143), so I added comprehensive coverage.

## My investigation process

I wrote a few scripts to systematically test functions with:
- Random invalid kwargs like `invalid_kwarg="test"`
- Common typos I've seen in practice (`dims` instead of `dim`, etc.)
- Both direct function calls and the `.azstats` accessor

All 24+ functions I tested handle invalid kwargs properly by raising TypeError. Great!

## What I added

**tests/test_kwargs_validation.py** - 26 tests organized by function category:
- Sampling diagnostics: ess, rhat, mcse, bfmi, diagnose
- Summary stats: summary, mean, median, mode, ci_in_rope  
- Visualization: qds, ecdf
- LOO/model comparison: loo, compare
- Metrics: bayesian_r2, metrics
- Other functions: thin, weight_predictions, bayes_factor, psense
- Accessor methods (6 tests for `.azstats.*` calls)

## Testing

```bash
pytest tests/test_kwargs_validation.py -v    # 26/26 passed
pytest                                        # 2475 passed, 2 skipped  
tox -e check                                  # all checks passed

### CODE REVIEW

File: tests/test_kwargs_validation.py [hunk 1]
```diff
@@ -0,0 +1,274 @@
...
+from arviz_base import load_arviz_data
+
+
+class TestKwargsValidationSamplingDiagnostics:
+    """Test kwargs validation for sampling diagnostic functions."""
+
+    @pytest.fixture(scope="class")
+    def idata(self):
+        """Load test data."""
+        return load_arviz_data("non_centered_eight")
```
OriolAbril: there is no need to define this fixture, the ones in `conftest.py` are available from any test file. For these tests you can use `datatree` directly

tests/test_kwargs_validation.py
```diff
@@ -0,0 +1,274 @@
...
+    def test_ess_rejects_invalid_kwargs(self, idata):
+        """Test that ess() raises TypeError for invalid kwargs."""
+        import arviz_stats as azs
+
+        with pytest.raises(TypeError, match=".*unexpected keyword argument.*"):
+            azs.ess(idata.posterior, invalid_kwarg="test")
+
+        # Also test common typos
+        with pytest.raises(TypeError, match=".*unexpected keyword argument.*"):
+            azs.ess(idata.posterior, methods="bulk")  # should be 'method'
```
OriolAbril: for testing functions a couple comments.

---

The first and main one is we want to test we don't silently ignore `**kwargs`. Therefore, these types of tests will only be relevant to functions that take `**kwargs`. There are many functions that do but not all of them do. [ess](https://python.arviz.org/projects/stats/en/latest/api/generated/arviz_stats.ess.html) in particular does not, so this test is testing basic builtin features of python functions. If we can't rely on that this test failing would be the least of our problems. Only functions that take `**kwargs` should get these kind of tests.

---

The second is about the test content and style so it is not relevant for ess, rhat, and other functions that do not take `**kwargs` but it is for all the ones who do. We ignore [doc related link checks]() in tests because the test name and body is usually self explanatory, we don't see any value added by adding a docstring that is basically the test name with better grammar. `arviz_stats` is the library we are testing so it should be imported only once at the top of the file with the other imports, and the two pytest.raises are doing the same check so one is enough, when it comes to checking if `invalid_kwarg` or `methods` is an argument of the function the function is doing exactly the same, we'd add the two different ones if we are triggering a different error or triggering the same error through a different code branch.

After doing all of that, the tests can be parametrized by the function to test on. As you probably have seen in https://github.com/arviz-devs/arviz-stats/pull/143, we have a single test function that checks multiple accessors through `pytest.mark.parametrize`. Here the same should happen with the relevant top level functions which I would rope together even if their source code is defined in different source files.

OriolAbril (commented):
Thanks this is a good start. I am starting with a few global comments and we can iterate from there. There might be some exceptions to the general comments I am mentioning but it will be easier to see those once the file is shorter and with less redundancy

tests/test_kwargs_validation.py
```diff
@@ -0,0 +1,64 @@
...
+        "mode",
+    ),
+)
+def test_accessor_kwargs_raise(datatree, func):
+    accessor = datatree.posterior.ds.azstats
+    with pytest.raises(TypeError, match=".*unexpected keyword argument.*"):
+        getattr(accessor, func)(invalid_kwarg="value")
+
+
+def test_loo_score_kwargs_raise(datatree):
```
OriolAbril: these bottom 3 can also be parametrized into one test function

OriolAbril (commented):
now that tests are parametrized and don't take as much space they should be moved (and combined/integrated with) the tests in `base/test_stats.py`. Otherwise we will be extremely confused in the future as to why the kde accessor is tested in one place but the ecdf one is tested somewhere else. Organization of tests here in arviz-stats is quite a disaster so we might also want to move them somewhere else which will be much easier if they are all in the same place

tests/base/test_stats.py
```diff
@@ -72,6 +72,46 @@ def test_extra_kwargs_raise(centered_eight, func):
         getattr(accessor, func)(dims="draw")
```
OriolAbril: these four cases should be merged into the test below. both are testing the same thing

OriolAbril (commented):
thanks! I think this will probably be the last round of review

tests/base/test_stats.py
```diff
@@ -64,12 +64,51 @@ def test_kde_is_normalized(bound_correction, kde_kwargs):
     assert_array_almost_equal(density_norm, 1, 6)
 
 
-@pytest.mark.parametrize("func", ("eti", "hdi", "kde", "histogram"))
-def test_extra_kwargs_raise(centered_eight, func):
-    accessor = centered_eight.posterior.ds.azstats
+@pytest.mark.parametrize(
+    ("func", "kwargs", "use_centered_eight"),
```
OriolAbril: for the same reasons explained in https://github.com/arviz-devs/arviz-stats/pull/314#discussion_r2884984298 there is no point in such a complicated parametrization. It doesn't matter if we use `dims` or `invalid_kwarg` nor it matters if we use `centered_eight` or `datatree` they are both datatrees with a posterior group and very similar variables and dimensions.

OriolAbril (changes requested):

Shlokpalrecha:
Updated per latest review: removed the complex kwargs/source parametrization, kept one simple test_accessor_kwargs_raise(datatree, func) plus the parametrized LOO test.

PR #319, opened 2026-03-09
fix: silence edge-case runtime warnings in dispersion helpers

sjiang83:
This small PR cleans up a few noisy NumPy warnings caused by edge-case inputs in the dispersion functions.

Instead of throwing warnings, _std and _var will now safely return np.nan for inputs smaller than 2, and _mad will do the same for empty arrays. I included some quick tests to verify this behavior.

This helps move us towards the GSoC DevOps goal of stricter warning handling, all while keeping the scope tightly contained.

Local testing commands:

## Testing

```bash
pytest -q tests/base/test_core.py
pytest -q tests/base/test_core.py -W error
pytest -q

### DIFF
--- a/tests/base/test_core.py
+++ b/tests/base/test_core.py
@@ -1,8 +1,6 @@
 """Tests for core base functions."""
 
 # pylint: disable=redefined-outer-name, no-self-use, protected-access
-import warnings
-
 import numpy as np
 import pytest
 from numpy.testing import assert_allclose, assert_array_almost_equal, assert_array_equal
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by sjiang83** [comment]
> * #### [docs/readthedocs.org:arviz-stats](https://arviz--319.org.readthedocs.build/projects/stats/en/319/) — Read the Docs build succeeded!

Thanks for the heads up！I fixed the formatting issue locally and pushed an updated commit. Please let me know if you still see any remaining linter problems！

File: tests/base/test_core.py [hunk 1]
```diff
@@ -315,6 +317,14 @@ def test_std_with_nan(self, core):
         result = core._std(x, skipna=True)
         assert not np.isnan(result)
 
+    def test_std_single_value_no_warning(self, core):
```
OriolAbril: this is doing the same as `test_std_single_value` and more so I don't see the point in keeping both. Given this is more extensive, the other one should be removed.

PR #320, opened 2026-03-09
Fix Numba array backend edge cases and maximize pure Python test coverage

player3zxcvbnm:
Hi,
I originally started looking into src/arviz_stats/numba/array.py just to bump up the test coverage, but I ended up stumbling onto a few edge-case bugs triggered by unhandled input types.
Most of the changes here are just handling None values properly (like testing axis=None in kde or bins=None in histogram). 
I noticed that all the remaining uncovered lines are strictly inside the guvectorize and jit blocks. Is it normal that pytest-cov can't see inside these compiled Numba functions, or is there a special way you usually test them?
Would really appreciate any feedback.

### DIFF
--- a/src/arviz_stats/numba/array.py
+++ b/src/arviz_stats/numba/array.py
@@ -72,12 +72,15 @@ class NumbaArray(BaseArray):
             axes = [(-1,), (0,), (0,)]
 
         quantile_ary = np.atleast_1d(quantile)
-
         # pylint: disable=no-value-for-parameter, unexpected-keyword-arg
         result = _quantile_ufunc(ary, quantile_ary, axes=axes)
+
+        result = np.moveaxis(result, 0, -1)
+
         if np.ndim(quantile) == 0:
-            return result
-        return np.moveaxis(result, 0, -1)
+            return result[..., 0]
+            
+        return result
 
     def _histogram(self, ary, bins=None, range=None, weights=None, density=True):  # pylint: disable=redefined-builtin
         """Compute the histogram of the data."""
--- a/src/arviz_stats/numba/array.py
+++ b/src/arviz_stats/numba/array.py
@@ -20,7 +20,7 @@ def process_ary_axes(ary, axes):
     reordered_axes = [i for i in range(ary.ndim) if i not in axes] + list(axes)
     ary = np.transpose(ary, axes=reordered_axes)
     ary = ary.reshape((*ary.shape[: -len(axes)], -1))
-    return ary, axes
+    return ary
 
 
 @guvectorize(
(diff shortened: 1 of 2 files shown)


tests/numba/test_array.py
```diff
@@ -139,26 +202,25 @@ def test_kde_basic(self, rng):
...
-    @pytest.mark.parametrize("axis", [0, 1, -1, -2])
-    def test_kde_axis(self, rng, axis):
+    @pytest.mark.parametrize(
+        "axis,expected_shape,expected_bw_shape",
+        [
+            (0, (20, 30, 128), (20, 30)),
+            (1, (10, 30, 128), (10, 30)),
+            (-1, (10, 20, 128), (10, 20)),
+            (-2, (10, 30, 128), (10, 30)),
+            (None, (128,), ()),
```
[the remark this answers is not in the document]
player3zxcvbnm: Thanks again for the review! Will   
take a look at it and push the updates soon.

OriolAbril (commented):
This looks good and we'll be able to merge once CI is all green. At the very least you'll need to run `tox -e check`, you might also need to rebase on `main`.

src/arviz_stats/numba/array.py
```diff
@@ -66,15 +66,22 @@ def quantile(self, ary, quantile, axis=-1, method="linear", skipna=False, weight
         axes = axis
         if axes is not None:
             ary, axes = process_ary_axes(ary, axes)
-            axes = [(-1,), (0,), (0,)]
+            axes = [axes, (0,), (0,)]
```
OriolAbril: This needs to be undone. I confused the process_ary_axes in the base module with this one. Here the axis we need to act on are moved to the end of the array and stacked together. We should go back to `(-1,)` and I would also update the `process_ary_axes` to not return the 2nd output because we don't use it anywhere and is only a source of confusion.

OriolAbril (commented):
As explained in https://python.arviz.org/projects/stats/en/latest/contributing/testing.html, the minimal and xarray testing jobs do not install numba, so the full job is the most relevant for this PR. You can see the failing tests at the bottom of the logs: https://github.com/arviz-devs/arviz-stats/actions/runs/24003327906/job/70004650837?pr=320#step:5:82839. In this case, the failures might be related to the incorrect suggestion I made about the `-1`/`axes` in axis-axes processing

src/arviz_stats/numba/array.py
player3zxcvbnm: Update: I realized that in my previous push I missed the tests in `test_array.py` and a few internal call sites (like in `kde`) that were still expecting the second return value.

I've just force-pushed a final clean version that fixes all those errors and verified it locally with `tox -r`.

OriolAbril (approved):

Issue #289, opened 2026-01-22
Add argument to `loo()` to pass log-likelihood function


jordandeklerk:
We do this for `loo_subsample()` in the large data scenario. But we should also probably do this for standard `loo()` for anyway working with very large data to avoid memory issues and just compute the log-likelihood on the fly.

See https://github.com/arviz-devs/arviz/issues/2197.

PR #322, opened 2026-03-10
FEAT: add custom log-likelihood to `loo()`

jordandeklerk:
Adds custom `log_lik_fn` support to `loo()`, similar to `loo_i()`. Users can pass a callable to compute log-likelihoods on the fly instead of requiring a pre-computed `log_likelihood` group. 

Wanted to call out that since `loo()` passes the full observed data and the unmodified `data` object to `log_lik_fn`, unlike `loo_i()` which passes a single observation, we don't need to align auxiliary groups (e.g., `constant_data`) to match the coords of the observation group via `_align_data_to_obs`. So it is a little easier here since we never have to worry about subsetting the observed data in anyway.

---
Resolves https://github.com/arviz-devs/arviz-stats/issues/289

### DIFF
--- a/src/arviz_stats/loo/loo.py
+++ b/src/arviz_stats/loo/loo.py
@@ -53,9 +53,8 @@ def loo(
     log_lik_fn : callable, optional
         Custom log-likelihood function. The signature must be ``log_lik_fn(observed, data)``
         where ``observed`` is an :class:`~xarray.DataArray` containing one or more observations
-        and ``data`` is the full :class:`~arviz_base.DataTree` or
-        :class:`~arviz_base.InferenceData`. The function must return an object with dimensions
-        ``("chain", "draw", *obs_dims)`` containing per-draw log-likelihood values.
+        and ``data`` is the full :class:`~xarray.DataTree`. The function must return an object
+        with dimensions ``("chain", "draw", *obs_dims)`` containing per-draw log-likelihood values.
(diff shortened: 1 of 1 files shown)


src/arviz_stats/loo/loo.py
```diff
@@ -49,6 +50,13 @@ def loo(
     reff : float, optional
         Relative MCMC efficiency, ``ess / n`` i.e. number of effective samples divided by the number
         of actual samples. Computed from trace by default.
+    log_lik_fn : callable, optional
+        Custom log-likelihood function. The signature must be ``log_lik_fn(observed, data)``
+        where ``observed`` is an :class:`~xarray.DataArray` containing one or more observations
+        and ``data`` is the full :class:`~arviz_base.DataTree` or
+        :class:`~arviz_base.InferenceData`. The function must return an object with dimensions
```
OriolAbril:
```suggestion
        and ``data`` is the full :class:`~xarray.DataTree`.
        The function must return an object with dimensions
```

if this is too dry or generic we could say "DataTree following the InferenceData schema"

OriolAbril (approved):
left a very minor doc related nit, everything looks great

src/arviz_stats/loo/loo.py
jordandeklerk: This is great

Issue #318, opened 2026-03-06
Improve docs for KDE


OriolAbril:
The docs for kde should mention where `**kwargs` of each function are forwarded to, and each of these functions where we forward things too should also appear in the docs (or alternatively, the whole of available kwargs hosuld be described in the first function without linking anywhere).

`arviz_stats.kde` mentions dataarray and array interfaces without linking: https://python.arviz.org/projects/stats/en/latest/api/generated/arviz_stats.kde.html#arviz_stats.kde.

`arviz_stats.base.dataarray_stats.kde` has no docstring at all so it need defining where kwargs go and add those functions also to `dataarray_stats` docs. (or say the description is in the array_stats docs and link there).

PR #327, opened 2026-03-12
DOC: Document kwargs for kde across all layers

player3zxcvbnm:
This PR resolves issue #318 by properly documenting the `**kwargs` passed to the `kde` function so users know they can use arguments like `bw` and `adaptive`.

**Note on scope:** I noticed that other functions in `accessors.py` (like `rhat`, `mcse`, `histogram`, etc.) should also be fixed. I kept this PR strictly scoped to `kde` , we could fix them in a separate PR for them!

### DIFF
--- a/src/arviz_stats/accessors.py
+++ b/src/arviz_stats/accessors.py
@@ -123,7 +123,7 @@ class _BaseAccessor:
         dim : str or sequence of str, optional
             Dimension(s) over which to compute the KDE.
         **kwargs : any, optional
-            Additional keyword arguments forwarded to the array or dataarray interface. 
+            Additional keyword arguments forwarded to the array or dataarray interface.
             See :func:`arviz_stats.base.array_stats.kde` for the full list of supported arguments.
 
         Returns
--- a/src/arviz_stats/accessors.py
+++ b/src/arviz_stats/accessors.py
@@ -124,7 +124,7 @@ class _BaseAccessor:
             Dimension(s) over which to compute the KDE.
         **kwargs : any, optional
             Additional keyword arguments forwarded to the array or dataarray interface. 
-            See the base ``kde`` for the full list of supported arguments.
+            See :func:`arviz_stats.base.array_stats.kde` for the full list of supported arguments.
 
         Returns
         -------
--- a/src/arviz_stats/accessors.py
+++ b/src/arviz_stats/accessors.py
@@ -116,7 +116,23 @@ class _BaseAccessor:
         return self._apply("mcse", sample_dims=sample_dims, method=method, prob=prob, **kwargs)
 
     def kde(self, dim=None, **kwargs):
-        """Compute the KDE for all variables in the dataset."""
+        """Compute the KDE for all variables in the dataset.
+
+        Parameters
+        ----------
+        dim : str or sequence of str, optional
+            Dimension(s) over which to compute the KDE.
+        **kwargs : dict, optional
+            Additional keyword arguments passed to the underlying KDE 
+            implementations. Supported arguments include `bw`, `adaptive`, 
+            `circular`, and `grid_len`. See the array-level `kde` methods 
+            for full details on these parameters.
+
+        Returns
+        -------
+        xarray.Dataset
+            A dataset containing the KDE grids and pdf values for each variable.
+        """
         return self._apply("kde", dim=dim, **kwargs)
 
     def qds(self, dim=None, **kwargs):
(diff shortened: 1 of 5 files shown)


src/arviz_stats/base/array.py
```diff
@@ -419,7 +419,7 @@ def get_bins(self, ary, axis=-1, bins="arviz"):
 
     # pylint: disable=redefined-builtin, too-many-return-statements
     # noqa: PLR0911
-    def histogram(self, ary, bins=None, range=None, weights=None, axis=-1, density=True):
+    def histogram(self, ary, bins=None, range=None, weights=None, axis=-1, density=None):
```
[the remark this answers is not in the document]
player3zxcvbnm: I've cleaned up the PR to remove the unrelated changes to `histogram`, and `quantile`. It should only contain the `kde` docstring updates now.

src/arviz_stats/base/array.py
```diff
@@ -429,7 +429,7 @@ def histogram(self, ary, bins=None, range=None, weights=None, axis=-1, density=T
         range : (float, float), optional
         weights : array-like, optional
         axis : int, sequence of int or None, default -1
-        density : bool, default True
+        density : bool, optional
```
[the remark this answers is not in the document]
player3zxcvbnm: Apologies for missing that one in the cleanup. I've just reverted the `density` parameter documentation to its original state as well. The PR should be completely clean now—containing only the `kde` docstring updates.

src/arviz_stats/accessors.py
```diff
@@ -116,7 +116,23 @@ def mcse(self, sample_dims=None, method="mean", prob=None, **kwargs):
         return self._apply("mcse", sample_dims=sample_dims, method=method, prob=prob, **kwargs)
 
     def kde(self, dim=None, **kwargs):
-        """Compute the KDE for all variables in the dataset."""
+        """Compute the KDE for all variables in the dataset.
```
aloctavodia, line 119: Do we want this plus a link to the main function or just a link to the main function?

src/arviz_stats/base/dataarray.py
```diff
@@ -228,7 +228,29 @@ def histogram(self, da, dim=None, bins=None, range=None, weights=None, density=T
...
+            Input data.
+        dim : str or sequence of str, optional
+            Dimension(s) over which to compute the KDE.
+        circular : bool, default False
+            Whether the data is circular (e.g., angles).
+        grid_len : int, default 512
+            Number of points on the KDE grid.
+        **kwargs : dict, optional
+            Additional keyword arguments passed to the array-level KDE
+            implementation. See the underlying array `kde` method for a
```
aloctavodia: This is linking to itself

player3zxcvbnm:
Thanks for the description edits and the feedback! I’m tied up for the next two days, but I'll address the linking and doc structure as soon as I'm back.

aloctavodia:
OK, let's go with the second option. But notice that we need to keep this "Forwarded to the array or dataarray interface for KDE." Also, notice that you are still missing kwargs, so double-check that.

aloctavodia:
Not clear why you change changed  `:func:` links to plain text (kde). Revert that change and also rebase.

aloctavodia:
The docs building is failing. Maybe you still need to rebase...

player3zxcvbnm:
Hi @aloctavodia, I’ve successfully rebased on main and fixed the trailing whitespace that the linter caught. But I noticed the `nightlies-xarray` CI check is failing. Is there anything i should do?

player3zxcvbnm:
Hi @aloctavodia, I retriggered the CI since I noticed that the issue seems to have been resolved, and now the CI checks have passed. Ready for review whenever you have a moment!

src/arviz_stats/base/array.py
```diff
@@ -573,12 +573,35 @@ def kde(self, ary, axis=-1, circular=False, grid_len=512, **kwargs):
...
+                If True, extends the observed range for linear KDE. Defaults to False.
+            * extend_fct : float, optional
+                Number of standard deviations to widen bounds if `extend` is True.
+                Defaults to 0.5.
+            * bound_correction : bool, optional
+                Whether to perform boundary correction on linear bounds. Defaults to True.
+            * custom_lims : list or tuple, optional
+                Custom bounds for the range of `ary`. Defaults to None.
+            * cumulative : bool, optional
+                If True, returns the cumulative PDF instead of the PDF. Defaults to False.
```
aloctavodia:
```suggestion
                If True, returns the CDF instead of the PDF. Defaults to False.
```

PR #344, opened 2026-03-29
Fix moment_match integration in loo

aatifjunaid:
- Ensures loo_moment_match is only called when moment_match=True
- Adds validation for required functions
- Adds minimal tests for flag behavior

All tests passing locally.

### DIFF
--- a/src/arviz_stats/loo/loo.py
+++ b/src/arviz_stats/loo/loo.py
@@ -2,6 +2,7 @@
 
 import warnings
 
+from .loo_moment_match import loo_moment_match
 from arviz_base import rcParams
 from xarray_einstats.stats import logsumexp
 
@@ -28,6 +29,11 @@ def loo(
     pareto_k=None,
     log_jacobian=None,
     mixture=False,
+    moment_match=False,
+    log_prob_upars_fn=None,
+    log_lik_i_upars_fn=None,
+    upars=None,
+    **moment_match_kwargs,
 ):
     r"""Compute Pareto-smoothed importance sampling leave-one-out cross-validation (PSIS-LOO-CV).
 
@@ -159,6 +165,8 @@ def loo(
     """
     loo_inputs = _prepare_loo_inputs(data, var_name)
     pointwise = rcParams["stats.ic_pointwise"] if pointwise is None else pointwise
+    if moment_match:
+        pointwise = True
 
     if reff is None:
         reff = _get_r_eff(data, loo_inputs.n_samples)
@@ -218,18 +226,35 @@ def loo(
             log_weights=mix_log_weights,
         )
 
-    return _compute_loo_results(
-        log_likelihood=loo_inputs.log_likelihood,
-        var_name=loo_inputs.var_name,
-        pointwise=pointwise,
-        sample_dims=loo_inputs.sample_dims,
-        n_samples=loo_inputs.n_samples,
-        n_data_points=loo_inputs.n_data_points,
-        log_weights=log_weights,
-        pareto_k=pareto_k,
-        approx_posterior=False,
-        log_jacobian=jacobian_da,
-    )
+    loo_res = _compute_loo_results(
+    log_likelihood=loo_inputs.log_likelihood,
+    var_name=loo_inputs.var_name,
+    pointwise=pointwise,
+    sample_dims=loo_inputs.sample_dims,
+    n_samples=loo_inputs.n_samples,
+    n_data_points=loo_inputs.n_data_points,
+    log_weights=log_weights,
+    pareto_k=pareto_k,
+    approx_posterior=False,
+    log_jacobian=jacobian_da,
+)
+    if moment_match:
+        if log_prob_upars_fn is None or log_lik_i_upars_fn is None:
+            raise ValueError(
+                "moment_match=True requires log_prob_upars_fn and log_lik_i_upars_fn"
+            )
+
+        loo_res = loo_moment_match (
+            data=data,
+            loo_orig=loo_res,
+            log_prob_upars_fn=log_prob_upars_fn,
+            log_lik_i_upars_fn=log_lik_i_upars_fn,
+            upars=upars,
+            var_name=var_name,
+            reff=reff,
+            **moment_match_kwargs,
+        )
+    return loo_res
 
 
 def loo_i(
--- a/tests/test_minimal.py
+++ b/tests/test_minimal.py
@@ -387,3 +387,24 @@ def test_hdi_invalid_prob(data_c0d1):
 def test_thin_invalid_factor(data_c0d1):
     with pytest.raises(ValueError, match="must be greater than 1"):
         array_stats.thin(data_c0d1, factor=0, chain_axis=0, draw_axis=1)
+
+
+def test_loo_moment_match_flag_runs():
+    from arviz_stats import loo
+    from arviz_base import load_arviz_data
+
+    data = load_arviz_data("centered_eight")
+
+    result = loo(data, moment_match=False)
+
+    assert result is not None
+
+def test_loo_moment_match_requires_functions():
+    from arviz_stats import loo
+    from arviz_base import load_arviz_data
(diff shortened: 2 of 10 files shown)


src/arviz_stats/loo/loo.py
jordandeklerk: Keep in mind that moment matching is a two-step algorithm. First compute standard PSIS-LOO-CV, then improve the estimates for observations with high Pareto-k. So `loo_moment_match()` expects a fully computed PSIS-LOO-CV result (with `pointwise=True` so it has per-observation Pareto k values) as its `loo_orig` argument so it knows which observations need fixing.

Since we're backend-agnostic and can't auto-derive the required functions from a model object the way R can with stanfit, `loo()` needs to handle this wiring explicitly. I think the cleanest way is to compute the base PSIS-LOO-CV via a recursive `loo()` call with `pointwise=True`, then pass that result directly to `loo_moment_match()` and return. 

So we should probably have something like the following at the entry point of this function:

```python
    if moment_match:
        # early validation goes here

        loo_orig = loo(
            data,
            pointwise=True,
            var_name=var_name,
            reff=reff,
            log_weights=log_weights,
            pareto_k=pareto_k,
            log_jacobian=log_jacobian,
        )

        return loo_moment_match(
            data,
            loo_orig,
            log_prob_upars_fn=log_prob_upars_fn,
            log_lik_i_upars_fn=log_lik_i_upars_fn,
            upars=upars,
            var_name=var_name,
            reff=reff,
            pointwise=pointwise,
            **moment_match_kwargs,
        )

    loo_inputs = _prepare_loo_inputs(data, var_name)
    pointwise = rcParams["stats.ic_pointwise"] if pointwise is None else pointwise
```

src/arviz_stats/loo/loo.py
```diff
@@ -164,8 +163,10 @@ def loo(
        Journal of Machine Learning Research, 25(72) (2024) https://jmlr.org/papers/v25/19-556.html
        arXiv preprint https://arxiv.org/abs/1507.02646
     """
-    loo_inputs = _prepare_loo_inputs(data, var_name, log_lik_fn=log_lik_fn)
+    loo_inputs = _prepare_loo_inputs(data, var_name)
     pointwise = rcParams["stats.ic_pointwise"] if pointwise is None else pointwise
+    if moment_match:
+        pointwise = True
```
jordandeklerk: This is correct that moment matching needs `pointwise=True` internally so the base PSIS-LOO-CV result includes per-observation Pareto k values to iterate over. But `loo_moment_match()` also accepts its own pointwise parameter which controls whether pointwise fields like `elpd_i`, `pareto_k`, and `influence_pareto_k` are included in the final result or stripped out before returning. By overwriting the variable here, the user's original choice is lost, so, for example, 

```python
loo(data, pointwise=False, moment_match=True)
```

would still return pointwise data.

I think we should remove these two lines since this should be handled with our original `loo` computations.

tests/test_minimal.py
jordandeklerk: There are two main problems here for these tests. Where the tests live, and what they test.

These tests should not be in `tests/test_minimal.py`. That file is for tests of the array-level API. The moment-match integration is a change to the `loo()` function, so the tests belong in `tests/loo/test_loo.py` alongside the existing PSIS-LOO tests. 

There's also `tests/loo/test_loo_mm.py` which already has the `roaches_r_example` fixture, callback functions, R reference values, and everything needed to write meaningful tests for moment matching. The new tests should reuse that existing infrastructure.

As a nudge toward the right direction for testing this new functionality, we want something like the following in spirit:

```python
@pytest.fixture(scope="module")
def roaches_r_example():
    from .test_loo_mm import load_roaches_r_example
    return load_roaches_r_example()


# Does the one-step path produce the same answer as the two-step path?
def test_loo_moment_match_matches_two_step(roaches_r_example):
    from arviz_stats import loo, loo_moment_match

    data = roaches_r_example["data_tree"]
    mm_kwargs = dict(
        log_prob_upars_fn=roaches_r_example["log_prob_fn"],
        log_lik_i_upars_fn=roaches_r_example["log_lik_i_fn"],
        upars=roaches_r_example["upars"],
        var_name="log_lik",
        cov=True,
    )

    loo_orig = loo(data, pointwise=True, var_name="log_lik")
    result_two_step = loo_moment_match(data, loo_orig, pointwise=True, **mm_kwargs)
    result_one_step = loo(data, moment_match=True, pointwise=True, **mm_kwargs)

    assert_almost_equal(result_one_step.elpd, result_two_step.elpd, decimal=10)
    assert_almost_equal(result_one_step.se, result_two_step.se, decimal=10)
    assert_almost_equal(result_one_step.p, result_two_step.p, decimal=10)
    assert_array_equal(result_one_step.pareto_k.values, result_two_step.pareto_k.values)
```

Also, as a general principle for testing, we should not just be testing that a result exists. This doesn't really tell us anything about the nature of the functionality and if it is robust or not. For some examples about what to be looking for in testing general PSIS-LOO-CV functions, see https://github.com/arviz-devs/arviz-stats/blob/main/tests/loo/test_loo.py.

tests/test_minimal.py
```diff
@@ -427,3 +387,24 @@ def test_hdi_invalid_prob(data_c0d1):
 def test_thin_invalid_factor(data_c0d1):
     with pytest.raises(ValueError, match="must be greater than 1"):
         array_stats.thin(data_c0d1, factor=0, chain_axis=0, draw_axis=1)
+
+
+def test_loo_moment_match_flag_runs():
+    from arviz_stats import loo
+    from arviz_base import load_arviz_data
```
jordandeklerk: Imports should not be nested inside the test functions. These should live at the top of the module.

tests/test_minimal.py
```diff
@@ -427,3 +387,24 @@ def test_hdi_invalid_prob(data_c0d1):
...
+
+    data = load_arviz_data("centered_eight")
+
+    result = loo(data, moment_match=False)
+
+    assert result is not None
```
jordandeklerk: This calls `loo(data, moment_match=False)` which is identical to `loo(data)`, so it doesn't actually test anything new. This should be removed.

tests/test_minimal.py
```diff
@@ -427,3 +387,24 @@ def test_hdi_invalid_prob(data_c0d1):
...
+
+def test_loo_moment_match_requires_functions():
```
jordandeklerk: Use the match parameter to verify the error message, e.g., 

```python
pytest.raises(ValueError, match="log_prob_upars_fn"))
```

jordandeklerk (changes requested):
Thanks for taking this on! You've made some good progress here and you've got the key pieces in place for the most part. That's the hardest part.

I'm taking a first pass at this and might come back with more changes after this first round. I will need to dive a little deeper and ensure everything is in the right place. The following items need to be addressed first though:

- There are some unrelated `histogram()` function changes on this branch that have nothing to do with the scope of this PR and the issue attached to it. All of these changes need to be reverted. I'm not sure where they came from. 

- One of the main things to address is the overall structure of how the moment match path fits into `loo()`. Right now the moment-match logic runs at the end of the function after the full PSIS-LOO-CV computation, which means validation happens too late, the pointwise flag gets permanently overridden, and the non-moment-match code path gets modified unnecessarily, among other things.

- The tests also need attention. They're currently in `tests/test_minimal.py` but should be in `tests/loo/test_loo.py` probably, and the existing `roaches_r_example` fixture from `tests/loo/test_loo_mm.py` should be reused to add end-to-end tests that prove the wiring is correct.

See the comments below for specifics on each point.

Remember to review the contributing tutorial here https://python.arviz.org/en/latest/contributing/pr_tutorial.html. You'll need to address linting issues and test CI failures as well. But we can iron those out with the new set of changes that need to be implemented.

jordandeklerk:
Closing this out to open a new PR after https://github.com/arviz-devs/arviz-stats/pull/373. This will add more complexity.

PR #350, opened 2026-04-05
Fix numba backend: handle scalar quantile and validate kde kwargs

aman-coder03:
## What this PR does
fixes two bugs in the `numba` backend that were causing 113 test failures when `stats.module` is set to `numba`

## Changes
1. Scalar quantile fix (`_quantile_ufunc`)
numba `_quantile_ufunc` gufunc has signature `(n),(m)->(m)` which requires the quantile argument to have at least 1 dimension. When a scalar quantile was passed (e.g. `0.5`), it would fail with,
```
ValueError: _quantile_ufunc: Input operand 1 does not have enough dimensions
```
**Fix:** wrap `quantile` with `np.atleast_1d()` before passing to the `ufunc` and squeeze the result back if the original input was scalar

2. kde kwargs validation
numba `kde` method accepted `**kwargs` and silently ignored unknown keyword arguments, while the base backend would raise a `TypeError`. This inconsistency caused `test_extra_kwargs_raise[kde]` to fail
**Fix:** explicitly validate kwargs at the start of the method and raise `TypeError` for any unrecognised ones, matching the behaviour of the base backend

## Testing
Ran the full arviz-stats test suite (2632 tests) with `stats.module : numba` set in `arvizrc`
`2632 passed, 2 skipped, 0 failed`
previously 113 failed

## Related
enables the `auto` option in `stats.module` to work correctly when `numba` is installed(arviz-devs/arviz-base#172)

### DIFF
--- a/src/arviz_stats/numba/array.py
+++ b/src/arviz_stats/numba/array.py
@@ -71,9 +71,11 @@ class NumbaArray(BaseArray):
             ary = ary.ravel()
 
         # pylint: disable=no-value-for-parameter, unexpected-keyword-arg
-        result = _quantile_ufunc(ary, quantile, axes=axes)
-        if np.ndim(quantile) == 0:
-            return result
+        scalar_quantile = np.ndim(quantile) == 0
+        quantile = np.atleast_1d(quantile)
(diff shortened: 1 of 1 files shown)


aman-coder03:
Closing this as it's been superseded by #320 which was merged yesterday and covers the same scalar quantile fix. Thanks for pointing it out @OriolAbril

PR #353, opened 2026-04-13
Add pareto_pit

aloctavodia:
The function `_pareto_pit`  computes PIT values using the ECDF, then refines the tails by fitting a generalized Pareto distribution (GPD) to the tail draws. This gives smoother, more accurate PIT values in the tails and avoids PIT values of 0 and 1. 

This follows the implementation in posterior package https://github.com/stan-dev/posterior/blob/727ac83cb52babb6eb54bdd1ce15f90439626945/R/pit.R#L238

### DIFF
--- a/src/arviz_stats/base/array.py
+++ b/src/arviz_stats/base/array.py
@@ -1240,6 +1240,7 @@ class BaseArray(_DensityBase, _DiagnosticsBase):
         chain_axis=-2,
         draw_axis=-1,
         random_state=None,
+        pareto_pit=False,
     ):
         """Compute LOO-PIT values with PSIS-LOO-CV weights.
 
@@ -1257,6 +1258,8 @@ class BaseArray(_DensityBase, _DiagnosticsBase):
             Axis for draws.
         random_state : int or Generator, optional
             Random seed or Generator for tie-breaking. If None, uses seed 214.
+        pareto_pit : bool, optional
+            If True, use Pareto-smoothed PIT values. Default is False.
 
         Returns
         -------
@@ -1275,7 +1278,7 @@ class BaseArray(_DensityBase, _DiagnosticsBase):
             rng = np.random.default_rng(random_state)
 
         loo_pit_ufunc = make_ufunc(self._loo_pit, n_output=1, n_input=3, n_dims=len(axes))
-        return loo_pit_ufunc(ary, y_obs, log_weights, rng=rng)
+        return loo_pit_ufunc(ary, y_obs, log_weights, rng=rng, pareto_pit=pareto_pit)
 
     def loo_expectation(
         self,
--- a/src/arviz_stats/base/array.py
+++ b/src/arviz_stats/base/array.py
@@ -1277,8 +1277,12 @@ class BaseArray(_DensityBase, _DiagnosticsBase):
         else:
             rng = np.random.default_rng(random_state)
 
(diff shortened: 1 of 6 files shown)


src/arviz_stats/base/diagnostics.py
```diff
@@ -932,6 +934,143 @@ def _summary(elpd_i, p_loo_i):
...
+        """Compute Pareto-smoothed PIT.
+
+        Compute PIT value using the ECDF, then refine in the tails by fitting a
+        generalized Pareto distribution (GPD) to the tail draws. This gives smoother, more
+        accurate PIT values in the tails, and avoids PIT values of 0 and 1.
+        The PIT values are not anymore rank based and continuous uniformity test is appropriate.
+
+        Parameters
+        ----------
+        draws_matrix : np.ndarray
```
OriolAbril:
```suggestion
        draws_matrix : np.ndarray of shape (n_obs, n_draws)
```

I would include this here already. This syntax is the common standard and will be recognized by docstub (and therefore included in the docstring once array type hints have good support for shape info)

src/arviz_stats/base/diagnostics.py
```diff
@@ -932,6 +934,143 @@ def _summary(elpd_i, p_loo_i):
...
+            2D array of posterior predictive draws with shape (n_obs, n_draws).
+        y_obs_array : np.ndarray
```
OriolAbril:
```suggestion
        y_obs_array : np.ndarray of shape (n_obs,)
```

src/arviz_stats/base/diagnostics.py
```diff
@@ -932,6 +934,143 @@ def _summary(elpd_i, p_loo_i):
...
+            1D array of observed values with shape (n_obs,).
+        log_weights : np.ndarray, optional
```
OriolAbril:
```suggestion
        log_weights : np.ndarray of shape (n_obs, n_draws) and dtype float, optional
```

not sure about adding the dtype too but it would be like this if we wanted to

src/arviz_stats/base/array.py
```diff
@@ -1274,8 +1277,12 @@ def loo_pit(
         else:
             rng = np.random.default_rng(random_state)
 
-        loo_pit_ufunc = make_ufunc(self._loo_pit, n_output=1, n_input=3, n_dims=len(axes))
-        return loo_pit_ufunc(ary, y_obs, log_weights, rng=rng)
+        n_sample_dims = len(axes)
+        sample_size = int(np.prod(ary.shape[-n_sample_dims:]))
+        ary = ary.reshape(*ary.shape[:-n_sample_dims], sample_size)
+        log_weights = log_weights.reshape(*log_weights.shape[:-n_sample_dims], sample_size)
```
OriolAbril, line 1283: not sure if we already have them but with all the reshapes in both here and `_loo_pit` method it would be great to make sure we have a test calling `loo_pit` (whichever entry point is fine as long as it is "above" this function) with uses >=2d array for `y_obs`. I am positive the code will already work but it is something easy to break without realizing if we don't have tests

aloctavodia: We have this https://github.com/arviz-devs/arviz-stats/blob/d34a84ffdca7d310d58ce6c3a0be430e83c59cdc/tests/loo/test_loo_pit.py#L100. Do you think that's good enough?

OriolAbril: Yes!

PR #358, opened 2026-04-29
Return shapley values in original order

aloctavodia:


jordandeklerk (approved):
Looks good as far as I can tell

PR #360, opened 2026-05-08
scale ISJ bw estimate by grid range rather than data range

zachjweiner:
The ISJ bandwidth is incorrectly rescaled from [0, 1] (over which `t_star` is estimated) back to the data domain - it should be rescaled by the width of the implemented grid (in data units) rather than just the range of the data. See https://github.com/tommyod/KDEpy/issues/199 for the brief discussion that led to this conclusion. The fix amounts to a wholesale replacement of `x_range` with `grid_range = grid_max - grid_min` in `density.py` (i.e., I checked that all instances should indeed be changed).

To further illustrate the improvement, ISJ bandwidth's variation with grid width, computed with
```python
import numpy as np
from arviz_stats.base import array_stats
from scipy.stats import norm
data = norm().rvs(10000)
np.array([array_stats.kde_linear(data, extend=True, bound_correction=False, extend_fct=f, bw="isj")[-1] for f in np.linspace(0, 2, 11)])
```
gives
```
array([0.17038331, 0.16231991, 0.15425661, 0.14671877, 0.13987671,
       0.13480469, 0.12944809, 0.12408305, 0.11988148, 0.11580617,
       0.11172394])
```
on main and
```
array([0.17310286, 0.17306596, 0.17305311, 0.17314323, 0.17296685,
       0.17303633, 0.17308081, 0.17317929, 0.17306257, 0.17254111,
       0.17325904])
```
on this PR, i.e., the result is much more stable to small variations in the grid width.

### DIFF
--- a/src/arviz_stats/base/density.py
+++ b/src/arviz_stats/base/density.py
@@ -74,7 +74,7 @@ class _DensityBase(_CoreBase):
         out = t - (2 * n * np.pi**0.5 * f) ** (-0.4)
         return out
 
-    def _root(self, function, n, args, x, x_range):
+    def _root(self, function, n, args, x, grid_range):
         # The right bound is at most 0.01
         found = False
         n = max(min(1050, n), 50)
@@ -89,7 +89,7 @@ class _DensityBase(_CoreBase):
                 tol *= 2.0
                 found = False
             if bw <= 0 or tol >= 1:
-                bw = (self.bw_silverman(x) / x_range) ** 2
+                bw = (self.bw_silverman(x) / grid_range) ** 2
                 return bw
         return bw
 
@@ -110,7 +110,7 @@ class _DensityBase(_CoreBase):
         bw = 0.9 * a * len(x) ** (-0.2)
         return bw
 
-    def bw_isj(self, x, grid_counts=None, x_std=None, x_range=None):
+    def bw_isj(self, x, grid_counts=None, x_std=None, grid_range=None):
         """Improved Sheather-Jones bandwidth estimation.
 
         Notes
@@ -126,17 +126,16 @@ class _DensityBase(_CoreBase):
         Ann. Statist. 38 (2010), no. 5, 2916--2957.
         """
         x_len = len(x)
-        if x_range is None:
-            x_min = np.min(x)
-            x_max = np.max(x)
-            x_range = x_max - x_min
 
         # Relative frequency per bin
         if grid_counts is None:
+            x_min = np.min(x)
+            x_max = np.max(x)
             x_std = np.std(x)
             grid_len = 256
             grid_min = x_min - 0.5 * x_std
             grid_max = x_max + 0.5 * x_std
+            grid_range = grid_max - grid_min
             grid_counts, _ = self._histogram(
                 x, bins=grid_len, range=(grid_min, grid_max), density=False
             )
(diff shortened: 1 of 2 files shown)


zachjweiner:
There's one other point of difference which doesn't usually matter in practice (with the notable exception of unthinned MCMC chains, i.e., with repeated samples due to rejected steps), which is the value of `n` used [here](https://github.com/arviz-devs/arviz-stats/blob/main/src/arviz_stats/base/density.py#L80). In KDEpy and Botev's [matlab implementation](https://www.mathworks.com/matlabcentral/fileexchange/14034-kernel-density-estimator/files/kde.m?s_tid=prof_contriblnk), it is set to the number of unique elements in `x` rather than the total length. Happy to add in this change, too.

tomicapretto:
Hi @zachjweiner, I had a look to the issue you opened in KDEpy as well as your contributions over there and here, and first of all thanks a lot for looking into it. 

It's been a while since I worked on the implementation of ISJ in ArviZ and honestly I didn't remember this detail.
I'm not sure I understand the stability numbers you show. Is it correct to say that with the new code `bw_fct` does not change the smoothness of the estimated density, and thus the estimated density at a point remains constant? The next question is: why would I want that?

On the other hand, I also had a look at Botev's code, and I'm seeing these two lines:

```matlab
% set up the grid over which the density estimate is computed;
R=MAX-MIN; dx=R/(n-1); xmesh=MIN+[0:dx:R]; N=length(unique(data));
```

and then:

```matlab
% take the rescaling of the data into account
bandwidth=sqrt(t_star)*R;
```

Am I missing anything? As far as I understand R is the range of the data.

zachjweiner:
Thanks for taking a look!

> As far as I understand R is the range of the data.

No, `R` is specifically the padded (grid) range. `Range` is the data range.

```matlab
minimum=min(data); maximum=max(data);
Range=maximum-minimum;
MIN=minimum-Range/2; MAX=maximum+Range/2;
```

> Is it correct to say that with the new code bw_fct does not change the smoothness of the estimated density, and thus the estimated density at a point remains constant? The next question is: why would I want that?

`bw_fct` is just a user option to scale the empirically determined bandwidth, right? I'm not seeing its relevance. I varied `extend_fct`, which just controls padding of the domain. I would think the bandwidth derived from the data distribution should be more or less invariant under varying padding.

tomicapretto:
@zachjweiner, I see what you mean now. I looked at things quickly initially, and wrongly understood that it used the bin width, not the grid length. To be clear, I see what you mean now and I think the fix is correct.

To get this merged, could you update the tests so they use the new argument?

@OriolAbril, @aloctavodia, I see the methods modify live within `_DensityBase`. Is it correct to say it is non user facing?

Thanks!!

aloctavodia:
@tomicapretto that's correct.

src/arviz_stats/base/density.py
```diff
@@ -110,7 +110,7 @@ def bw_silverman(self, x, x_std=None, **kwargs):  # pylint: disable=unused-argum
         bw = 0.9 * a * len(x) ** (-0.2)
         return bw
 
-    def bw_isj(self, x, grid_counts=None, x_std=None, x_range=None):
+    def bw_isj(self, x, grid_counts=None, x_std=None, grid_range=None):
```
tomicapretto, line 113: @zachjweiner, sorry I didn't see this earlier. But with the introduced change there's the possibility that `grid_range` is `None` while `grid_counts` is `None` as well, which would introduce an error in `self._root(...)`. I think we need to first check if `grid_range` is `None`, and if that happens we compute it. The `if grid_counts is None` block would be similar, but without the computation of the `grid_range` (as it will be either passed or computed before).

tomicapretto (changes requested):

src/arviz_stats/base/density.py
zachjweiner: I noticed that, but I disregarded it because the existing code seemed designed only for the extra arguments to either be all None or all precomputed (and it seems to only be used in these two forms). E.g., the previous version doesn't work if x_range is passed but grid_counts is not, because the `grid_counts is None` block assumed x_min and x_max are in the namespace (via the `x_range is None` block). That's actually why I merged them - the grid_range calculation, unlike x_range, relies on x_std as well, and I didn't want to introduce needless recomputation in the "all None" case.

I'm happy to change this as you see fit for this PR. I just wasn't sure these options were user-facing, and I thought this logic would merit broader attention throughout _DensityBase if one wanted to support mixed argument cases.

tomicapretto: You're right these are not user facing. Let's leave things as they are and only change them if we decide we to modify the module as a whole for this kind of consistency.

tomicapretto (approved):

PR #368, opened 2026-06-17
fix: convert ary to mutable numpy array in _ps_tail to support JAX arrays

### LINKED ISSUE #367: az.loo() crashes with JAX installed: immutable array in _ps_tail
## Bug

`az.loo()` raises a `TypeError` when JAX is installed alongside arviz-stats, because `_ps_tail` in `arviz_stats/base/diagnostics.py` performs an in-place item assignment on an array that xarray has backed with JAX under the hood.

JAX arrays are immutable: `ary[idx] = val` raises:

```
TypeError: JAX arrays are immutable and do not support in-place item assignment.
Instead of x[idx] = y, use x = x.at[idx].set(y) ...
```

## Environment

```
arviz >= 1.1.0
arviz-stats >= 1.1.0
jax >= 0.4
```

## Minimum Working Example

No numpyro required — any InferenceData whose log-likelihood values are JAX arrays reproduces the crash:

```python
import numpy as np
import jax.numpy as jnp
import arviz as az
from arviz_base import from_dict

rng = np.random.default_rng(0)
nchains, ndraws, nobs = 4, 500, 8

idata = from_dict(
    {
        "posterior": {"mu": rng.normal(size=(nchains, ndraws))},
        "log_likelihood": {
            # Wrap log-likelihood in a JAX array to simulate a JAX backend
            "obs": jnp.array(rng.normal(size=(nchains, ndraws, nobs)))
        },
        "observed_data": {"obs": rng.normal(size=nobs)},
    },
    dims={"obs": ["obs_dim"], "log_likelihood": ["obs_dim"]},
)

# Crashes with TypeError: JAX arrays are immutable ...
loo = az.loo(idata, pointwise=True)
print(loo)
```

## Full Traceback

```
Traceback (most recent call last):
  File "mwe.py", line 20, in <module>
    loo = az.loo(idata, pointwise=True)
  ...
  File ".../arviz_stats/base/diagnostics.py", line 1210, in _ps_tail
    ary[ordered[tail_ids]] = smoothed
TypeError: JAX arrays are immutable and do not support in-place item assignment.
Instead of x[idx] = y, use x = x.at[idx].set(y) or another .at[] method:
https://docs.jax.dev/en/latest/_autosummary/jax.numpy.ndarray.at.html
```

## Root Cause

`_ps_tail` (line ~1210 of `diagnostics.py`) does:

```python
ary[ordered[tail_ids]] = smoothed
```

When JAX is installed, xarray's `apply_ufunc` can dispatch array operations through JAX's numpy backend, so `ary` is a JAX array by the time it reaches `_ps_tail`. JAX arrays do not support in-place mutation.

## Proposed Fix

At the top of `_ps_tail`, force `ary` to a mutable numpy array before any operations:

```python
# JAX arrays are immutable; copy to a plain numpy array so in-place assignments work.
ary = np.asarray(ary).copy()
```

This is safe because `_ps_tail` is a pure-numpy internal helper that should never be operating on JAX arrays in the first place. `numpy` is already imported as `np` in the file.

TheSkyentist:
Fixes #367.

## Summary

- `_ps_tail` performs an in-place assignment (`ary[ordered[tail_ids]] = smoothed`) that crashes when `ary` is a JAX array, because JAX arrays are immutable.
- Add `ary = np.asarray(ary).copy()` at the top of `_ps_tail` to ensure it always works on a mutable numpy array before doing any operations.
- Add a regression test in `tests/loo/test_loo.py` that passes JAX arrays as the log-likelihood values and asserts `loo()` completes without error.

## Test plan

- [ ] `test_loo_jax_arrays` — new test, passes a `jnp.array`-backed log-likelihood group through `loo()` and checks it doesn't raise and returns finite values. Skipped automatically if JAX is not installed.
- [ ] Full `tests/loo/test_loo.py` suite — all 30 existing tests still pass.

### DIFF
--- a/src/arviz_stats/base/diagnostics.py
+++ b/src/arviz_stats/base/diagnostics.py
@@ -1164,6 +1164,9 @@ class _DiagnosticsBase(_CoreBase):
         k : float
             Estimated shape parameter.
         """
+        # JAX arrays are immutable; copy to a plain numpy array so in-place assignments work.
+        ary = np.asarray(ary).copy()
+
         if log_weights:
             ary = ary - np.max(ary)
 
--- a/tests/loo/test_loo.py
+++ b/tests/loo/test_loo.py
@@ -554,3 +554,20 @@ def test_loo_mixture(centered_eight):
     assert np.isfinite(result_mix.elpd)
     assert np.all(np.isfinite(result_mix.elpd_i.values))
     assert np.all(np.isfinite(result_mix.pareto_k.values))
+
+
+def test_loo_jax_arrays(centered_eight):
+    """loo() must not raise when log_likelihood contains JAX arrays (immutability regression)."""
+    jnp = importorskip("jax.numpy")
+
+    # Copy first: centered_eight is a session-scoped fixture and must not be mutated.
+    data = centered_eight.copy()
+    # Replace the log_likelihood group values with JAX arrays to simulate a JAX backend.
+    obs = data.log_likelihood["obs"]
+    data.log_likelihood["obs"] = xr.DataArray(
+        jnp.asarray(obs.values), dims=obs.dims, coords=obs.coords
+    )
+
+    result = loo(data, pointwise=True)
+    assert np.isfinite(result.elpd)
+    assert np.all(np.isfinite(result.pareto_k.values))
--- a/tests/loo/test_loo_moment_matching_helpers.py
+++ b/tests/loo/test_loo_moment_matching_helpers.py
@@ -142,6 +142,9 @@ class TestMomentMatchFunctions:
             draws=50,
             tune=50,
             chains=2,
+            # sample sequentially, forking workers triggers a RuntimeWarning
+            # from JAX's at-fork handler that our filterwarnings turns into an error
+            cores=1,
             random_seed=3,
             progressbar=False,
         )
--- a/tox.ini
+++ b/tox.ini
@@ -47,6 +47,7 @@ extras =
     numba
 deps =
     pymc>=6
+    jax
 setenv =
     ARVIZ_REQUIRE_ALL_DEPS=TRUE
 


OriolAbril (commented):
Thanks for the fix. Give me a couple days to discuss how to properly incorporate the test and get back. We have multiple envs running the tests on CI but none of them have jax so far. We'll probably want to add it to the env that has everything and use the custom importorskip so tests fail when running from that env if jax is actually missing.

TheSkyentist:
Following up here. The test already uses the repo's `importorskip` helper, which raises instead of skipping when `ARVIZ_REQUIRE_ALL_DEPS` is set, so that part matches what you described. The piece that was missing is that `jax` was never added to the `full` tox env, so CI on this PR has been failing with `ModuleNotFoundError: No module named 'jax'` on the `full` job since your comment.

I just pushed a small change adding `jax` as a dependency of `[testenv:full]` in `tox.ini`, so the env that sets `ARVIZ_REQUIRE_ALL_DEPS=TRUE` now actually has jax installed and the test runs for real instead of raising on import. Let me know if that matches what you had in mind, or if you'd rather scope it differently, for example a dedicated `jax` extra in `pyproject.toml`.

OriolAbril (approved):
looks good, thanks! And sorry for the slow reviews. Should be good to merge once you rebase/merge with main

TheSkyentist:
I've rebased onto current main.

I also found a bug in my tests, I was mutating session-scoped `centered_eight` in `test_loo_jax_arrays`, so every subsequent test got the polluted version. It didn't show up before purely because of test ordering. The test now copies first and preserves the coords, which matches the centered_eight_with_sigma pattern already in tests/conftest.py. Slight change to the original fix commit.

TheSkyentist:
Also, I believe the test failures are also failures on main unrelated to these code changes.

jordandeklerk:
> Also, I believe the test failures are also failures on main unrelated to these code changes.

As far as I can see, these changes here, specifically adding Jax as a testing dependency, is the cause of these failures. 

It looks like `test_log_lik_i_matches_arviz_log_likelihood_group` calls `pm.sample` which forks worker processes so Jax's at-fork handler now fires a "fork is incompatible with multithreaded code" `RuntimeWarning` that `filterwarnings = ["error"]` escalates into a test failure. On main, Jax is never imported so no handler is registered. 

It looks like Python 3.14 passes because 3.14 already defaults to `forkserver` instead of `fork` https://docs.python.org/3/library/multiprocessing.html#contexts-and-start-methods.

---

So I believe adding a simple `cores=1` to this test (https://github.com/arviz-devs/arviz-stats/blob/main/tests/loo/test_loo_moment_matching_helpers.py#L140) in the `pm.Sample()` block fixes this, but would want to test this more. 

@OriolAbril Any thoughts on this, or a better approach?

OriolAbril:
I think the cores alternative sounds good @jordandeklerk, if we see it slows down testing a bit (even if only sampling for a handful of iterations) we could also try circumventing `pm.sample` directly and call `pm.compute_log_likelihood` using a synthetic inferencedata with the expected groups and variables

aloctavodia:
Is this ready @jordandeklerk to merge?

jordandeklerk:
> Is this ready @jordandeklerk to merge?

LGTM!

PR #370, opened 2026-06-23
fix(summary): preserve numeric dtypes by removing string formatting in _round_summary

Vitaliy-Pikalo:
## Summary

Fixes #[arviz-devs/arviz#2591](https://github.com/arviz-devs/arviz/issues/2591) — `az.summary()` returns `object` dtype columns when `round_to="auto"`, breaking downstream numeric operations like `.mean()`, comparisons, and plotting.

## Root cause

`_round_summary` in `summary.py` had a Rule 5 block that applied Python f-string formatting to all float columns before returning:

```python
# Rule 5 (old) — converted floats to strings
result[col] = result[col].apply(lambda x: f"{x:.2f}" if np.isfinite(x) else x)
# or
result[col] = [f"{val:g}" or f"{val:.0e}" for ...]
```

Once any cell in a column becomes a Python string, pandas promotes the entire column dtype to `object`. This affected every float column in the summary output.

## Fix

- Removed the `use_scientific` dict and the scientific-notation branching in Rule 3
- Removed Rule 5 entirely — values stay as native Python floats
- Rule 3 now always calls `round_num(stat_val, decimal_places)` without conditionally converting to strings

## Before / after

```python
# Before
summary = az.summary(idata)
print(summary.dtypes)
# mean      object  ← broken
# sd        object
# hdi_3%    object

# After
summary = az.summary(idata)
print(summary.dtypes)
# mean      float64  ← correct
# sd        float64
# hdi_3%    float64
```

## Notes

Display formatting (scientific notation, significant figures) was the original intent of Rule 5, but that concern belongs at the render layer (e.g. `pd.options.display.float_format`), not in the data layer. The numeric values themselves are now correctly rounded via `round_num` and stay as floats.

### DIFF
--- a/src/arviz_stats/summary.py
+++ b/src/arviz_stats/summary.py
@@ -275,7 +275,6 @@ def _round_summary(summary_result, round_val):
     result = summary_result.copy()
     columns = result.columns
     rounded_columns = set()
-    use_scientific = {}
 
     # Rule 1: ESS columns and min_ss are rounded down to int
     ess_cols = [col for col in columns if col.startswith("ess_") or col == "min_ss"]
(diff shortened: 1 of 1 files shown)


aloctavodia:
Thanks for taking the time to write this PR. The behaviour of "auto" is intentional and documented in the docstring and there is no interest in changing it, so I am closing this issue.

PR #371, opened 2026-07-01
fix: standardize loo_influence for std/var by dividing by the statistic itself (#337)

### LINKED ISSUE #337: standardization for loo influence for std/var
standardization in `loo_influence` seems to be to make the influence a dimensionless quantity, so that if a parameter undergoes an affine transformation the influence would remain unchanged. For `mean` it standardizes by `sd` and for `median` it standardizes by `mad`; all 4 of these quantities share the same dimension (i.e. if we scale the data by `b`, all 4 quantities will scale by `b`, while `mean` and `median` will shift):

https://github.com/arviz-devs/arviz-stats/blob/3c3c8a8d6cc402bf5a0994f59e89be1811d1b31c/src/arviz_stats/loo/loo_influence.py#L155-L158

For `sd` and `var` no standardization is performed, but each of the corresponding influences could be made dimensionless by scaling by the metric itself, i.e. just change the default
https://github.com/arviz-devs/arviz-stats/blob/3c3c8a8d6cc402bf5a0994f59e89be1811d1b31c/src/arviz_stats/loo/loo_influence.py#L138
to
```python
func_s = func
```

laishettikarthik-tech:
Closes #337

## What this PR does
For `kind="std"` and `kind="var"`, `loo_influence` with `standardize=True` 
previously left `func_s = None` and skipped the standardization step, 
returning a quantity with units rather than a dimensionless influence.

Two changes:
- `func_s = func` (was `None`) so the same function used to compute the 
  statistic is used for standardization
- Extended the `if standardize` condition to include `"std"` and `"var"` 
  alongside the existing `"mean"` and `"median"`

## Math
- `std`: shift/std is dimensionless — scaling data by b scales both by b
- `var`: shift/var is dimensionless — scaling data by b scales both by b²

This matches the existing pattern for mean (standardized by sd) and 
median (standardized by mad).

## Tests
Added `test_loo_influence_standardize_scale_invariant` parametrized over 
`["std", "var"]` — verifies that scaling all data by a constant leaves 
the standardized influence unchanged.

### DIFF
--- a/src/arviz_stats/loo/loo_influence.py
+++ b/src/arviz_stats/loo/loo_influence.py
@@ -144,17 +144,17 @@ def loo_influence(
             func_s = mad
         elif kind == "sd":
             func = std
+            func_s = func
         elif kind == "var":
             func = var
+            func_s = func
 
         shift = np.abs(
             loo_expec
             - func(data, group=group, var_names=var_names, dim=sample_dims, round_to="none").dataset
         )
 
-        if standardize and kind in ["mean", "median", "std", "var"]:
+        if standardize and kind in ["mean", "median", "sd", "var"]:
             shift /= func_s(
                 data, group=group, var_names=var_names, dim=sample_dims, round_to="none"
             ).dataset
-
-    return shift, khat
--- a/src/arviz_stats/loo/loo_influence.py
+++ b/src/arviz_stats/loo/loo_influence.py
@@ -158,3 +158,5 @@ def loo_influence(
             shift /= func_s(
                 data, group=group, var_names=var_names, dim=sample_dims, round_to="none"
             ).dataset
+
+    return shift, khat
--- a/src/arviz_stats/loo/loo_influence.py
+++ b/src/arviz_stats/loo/loo_influence.py
@@ -135,7 +135,7 @@ def loo_influence(
         )
 
         func = None
-        func_s = func
+        func_s = None
         if kind == "mean":
             func = mean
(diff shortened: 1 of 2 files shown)


aloctavodia:
@laishettikarthik-tech, thanks for your help. Please double-check your changes, including tests, and make sure they pass (including linters)

src/arviz_stats/loo/loo_influence.py
```diff
@@ -135,7 +135,7 @@ def loo_influence(
         )
 
         func = None
-        func_s = None
+        func_s = func
```
sethaxen: Now that `func_s` is set below for every `kind`, we should just leave this as `None`:
```suggestion
        func_s = None
```

sethaxen (changes requested):
Thanks for the PR! The approach looks good to me. One minor request.

I think all test failures are unrelated and fixed by #377. Perhaps you could add an entry to the changelog?

laishettikarthik-tech:
Thanks for the review! I've applied the func_s = None fix. On the changelog — I looked and CHANGELOG.md looks auto-generated from GitHub Releases via the post-release workflow, so I don't see a manual entry point for individual PRs. Could you point me to what you mean, or is there a different convention (PR title, label, etc.) I should follow instead?

sethaxen (approved):
Ah, of course, forgot it's autogenerated now. Sorry about that!

LGTM, thanks again!

PR #373, opened 2026-07-02
Add helper functions for moment matching with PyMC and Bambi

aloctavodia:


src/arviz_stats/loo/loo_moment_match.py
```diff
@@ -165,6 +172,27 @@ def loo_moment_match(
             "Please compute the initial LOO with pointwise=True."
         )
 
+    # Auto-build moment matching functions from model if provided
+    if model is not None and (
+        log_prob_upars_fn is None or log_lik_i_upars_fn is None or upars is None
+    ):
+        from arviz_stats.loo.loo_moment_match_helper import mm_from_pymc
+
+        ## if model is Bambis's, apply moment match correction to data
```
aloctavodia: This is something that needs to be added to Bambi.

jordandeklerk:
Very exciting! Will take a brief pass today.

src/arviz_stats/loo/loo_moment_match_helper.py
```diff
@@ -0,0 +1,317 @@
...
+        Computes the log likelihood of a single observation ``i`` (a flat
+        integer index in C-order over the observation dimensions of the
+        selected observed variable).
+    upars : xarray.DataArray
+        Posterior draws transformed to the unconstrained parameter space, with
+        dims ``(chain, draw, unconstrained_parameter)`` and coordinate values
+        given by the (possibly transformed) value-variable names (with
+        ``[i,j,...]`` suffixes for multi-element value variables).
+    """
+    from pymc.model import modelcontext
```
jordandeklerk, line 56: Can we move this to the top with other imports?

src/arviz_stats/loo/loo_moment_match_helper.py
```diff
@@ -0,0 +1,317 @@
...
+
+    Returns
+    -------
+    upars : DataArray
+        Shape ``(chain, draw, n_params)``.
+    n_params : int
+        Total number of unconstrained parameters.
+    """
+    import pytensor
+    import pytensor.tensor as pt
```
jordandeklerk, line 219: Same as above

src/arviz_stats/loo/loo_moment_match_helper.py
```diff
@@ -0,0 +1,317 @@
...
+    if model.discrete_value_vars:
+        raise NotImplementedError(
+            "Moment matching is not supported for models with discrete random"
+            f"Found discrete value variables: {[v.name for v in model.discrete_value_vars]}."
+        )
+    if model.potentials:
+        raise NotImplementedError(
+            "Moment matching for models with potentials is not supported"
+            "Remove the potentials or supply the moment matching functions manually."
+        )
```
jordandeklerk, line 143: nit: there's no separator between these strings so they got lumped together in the error message.


```suggestion
    if model.discrete_value_vars:
        raise NotImplementedError(
            "Moment matching is not supported for models with discrete random variables. "
            f"Found discrete value variables: {[v.name for v in model.discrete_value_vars]}."
        )
    if model.potentials:
        raise NotImplementedError(
            "Moment matching for models with potentials is not supported. "
            "Remove the potentials or supply the moment matching functions manually."
        )
```

src/arviz_stats/loo/loo_moment_match_helper.py
```diff
@@ -0,0 +1,317 @@
...
+        Posterior draws transformed to the unconstrained parameter space, with
+        dims ``(chain, draw, unconstrained_parameter)`` and coordinate values
+        given by the (possibly transformed) value-variable names (with
+        ``[i,j,...]`` suffixes for multi-element value variables).
+    """
+    from pymc.model import modelcontext
+
+    model = modelcontext(model)
+    if model is None:
+        raise ValueError("A PyMC model is required to build moment matching functions.")
```
jordandeklerk: I think `modelcontext(model)` raises `TypeError ("No model on context stack.")` when there is no model and no active model context, so the `if model is None: raise ValueError(...)` right below it is never reached and users see the PyMC error.

I think this is okay


```suggestion
    try:
        model = modelcontext(model)
    except TypeError as err:
        raise ValueError(
            "A PyMC model is required to build moment matching functions."
        ) from err
```

src/arviz_stats/loo/loo_moment_match_helper.py
```diff
@@ -0,0 +1,317 @@
...
+__all__ = ["mm_from_pymc"]
+
+
+def mm_from_pymc(
+    idata,
+    *,
+    model=None,
+    var_name=None,
+):
+    """Build the functions and posterior array needed by loo_moment_match from a PyMC model.
```
jordandeklerk:
```suggestion
    """Build the functions and posterior array needed by :func:`arviz_stats.loo_moment_match` from a PyMC model.
```

src/arviz_stats/loo/loo_moment_match_helper.py
```diff
@@ -0,0 +1,317 @@
...
+    joined_unc = pt.concatenate([pt.reshape(b, (pt.shape(b)[0], -1)) for b in new_outputs], axis=1)
+    transform_func = pytensor.function([joined], joined_unc, on_unused_input="ignore")
+
+    # Assemble the input matrix from posterior draws and apply the transform
+    flat_blocks = []
+    for post_name, unc_shape in zip(posterior_names, unc_shapes):
+        alen = int(np.prod(unc_shape, dtype=int)) if unc_shape else 1
+        flat_blocks.append(
+            np.asarray(idata.posterior[post_name].values).reshape(chain * draw, alen)
+        )
```
jordandeklerk: I think this crashes for transforms where the constrained and unconstrained parameter counts differ (e.g. Dirichlet or LKJ Cholesky). `unc_shape` is the unconstrained shape (from `initial_point`, which is already in the transformed space), but it is used both to slice the batch-input columns for the constrained variable `cvar = rv.type()` and to reshape the constrained posterior draws `idata.posterior[rv.name]`. 

For a 3-simplex, for example, the posterior is (chain, draw, 3) while `unc_shape` is (2,), so the reshape raises an error. Tracking the input (constrained) shape separately and using it for the slice and the posterior reshape fixes this, I believe, while keeping `unc_shape` for the output layout and labels.

I think something like below works, but probably worth testing a bit more just to make sure.


```suggestion
    posterior_names = []
    unc_shapes = []  # unconstrained shapes (from initial_point)
    in_shapes = []   # input (posterior/constrained) shapes

    for v in value_vars:
        rv = model.values_to_rvs[v]
        transform = model.rvs_to_transforms[rv]
        unc_shape = initial_point[v.name].shape

        if transform is not None and rv.name in posterior_data_vars:
            cvar = rv.type()
            cvar.name = rv.name
            inputs_for_func.append(cvar)
            outputs_for_func.append(transform.forward(cvar, *rv.owner.inputs))
            posterior_names.append(rv.name)
            in_shapes.append(tuple(idata.posterior[rv.name].shape[2:]))
        else:
            post_name = rv.name if rv.name in posterior_data_vars else v.name
            if post_name not in posterior_data_vars:
                raise KeyError(
                    f"Could not find {v.name!r} or {rv.name!r} in idata.posterior. "
                    f"Available: {list(posterior_data_vars)}"
                )
            inputs_for_func.append(v)
            outputs_for_func.append(v)
            posterior_names.append(post_name)
            in_shapes.append(unc_shape)

        unc_shapes.append(unc_shape)

    joined = pt.matrix("upars_input", dtype="float64")
    replace = {}
    last = 0
    
    for cvar, in_shape in zip(inputs_for_func, in_shapes):
        alen = int(np.prod(in_shape, dtype=int)) if in_shape else 1
        replace[cvar] = joined[:, last : last + alen].reshape((joined.shape[0], *in_shape))
        last += alen
        
    new_outputs = pytensor.graph.vectorize_graph(outputs_for_func, replace)
    joined_unc = pt.concatenate([pt.reshape(b, (pt.shape(b)[0], -1)) for b in new_outputs], axis=1)
    transform_func = pytensor.function([joined], joined_unc, on_unused_input="ignore")

    flat_blocks = []
    for post_name, in_shape in zip(posterior_names, in_shapes):
        alen = int(np.prod(in_shape, dtype=int)) if in_shape else 1
        flat_blocks.append(
            np.asarray(idata.posterior[post_name].values).reshape(chain * draw, alen)
        )
```

jordandeklerk (approved):
Thanks for this! Very excited to use this and test it out. 

There is one issue and a few minor ones that I think we should address before merging. I think the major one is that `mm_from_pymc` fails for any model whose constrained and unconstrained parameter counts differ.

For example, running this simple Dirichlet model failed for me with the `ValueError`. Maybe something on my end, but worth looking into I think.

```python
y = np.array([12, 8, 5])
with pm.Model() as model:
    theta = pm.Dirichlet("theta", a=np.ones(3))
    obs = pm.Multinomial("obs", n=int(y.sum()), p=theta, observed=y)
    idata = pm.sample(draws=200, tune=200, chains=2)

mm_from_pymc(idata, model=model)
# ValueError: cannot reshape array of size 1200 into shape (400, 2)
```

---

Also I think the CI errors are fised with a rebase I would imagine.

src/arviz_stats/loo/loo_moment_match_helper.py
aloctavodia: Don't we need to keep the imports inside the function so pymc is not imported unless it is needed?

jordandeklerk: Oh yeah good point. Forgot we won't have pymc as a dependency.

jordandeklerk:
Made some additional changes here. Let me know what you think.

src/arviz_stats/loo/loo_moment_match.py
```diff
@@ -165,6 +172,33 @@ def loo_moment_match(
...
+        if hasattr(model, "moment_match_correction"):
+            data = model.moment_match_correction(data)
```
aloctavodia:
```suggestion
            data = model._re_center_intercept(data)
```

aloctavodia: There is an open PR, but not finished yet https://github.com/bambinos/bambi/pull/994

tox.ini
```diff
@@ -45,6 +45,8 @@ extras =
     test-xarray
     xarray
     numba
+deps =
+    pymc>=5
```
aloctavodia: pymc>=6

aloctavodia (commented):
LGTM

PR #375, opened 2026-07-03
REF: refactor `loo_kfold` to fix group assignment

jordandeklerk:
Fixes grouped k-fold split to always produce exactly k folds.

`loo_kfold(..., group_by=...)` (via `_kfold_split_grouped`) produced the wrong number of folds whenever the number of groups was not an exact multiple of `k`. Instead of `k` folds it returned fewer, with one or more empty folds, and for cases like an odd number of groups with `k=2`, it collapsed every group into a single fold. Downstream, an empty fold gets fed into the `elpd` computation and crashes `loo_kfold` with an `IndexError`. Since `k` defaults to 10, essentially any grouped dataset whose group count isn't a multiple of `k` was affected.

### DIFF
--- a/tests/loo/test_helper_kfold.py
+++ b/tests/loo/test_helper_kfold.py
@@ -171,8 +171,7 @@ def test_kfold_split_grouped_many_groups():
     x_many_groups = np.repeat(np.arange(10), 3)
     k = 3
     folds = _kfold_split_grouped(k=k, x=x_many_groups)
-    assert len(set(folds)) <= k
-    assert len(set(folds)) >= 1
+    assert set(folds.tolist()) == set(range(1, k + 1))
     assert min(folds) >= 1
     assert max(folds) <= k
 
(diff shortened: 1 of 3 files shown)


tests/loo/test_helper_kfold.py
```diff
@@ -483,3 +482,47 @@ def test_get_fold_indices_unequal_folds():
...
+        (11, 3),
+    ],
+)
+def test_kfold_split_grouped_balanced_fold_sizes(n_levels, k):
+    x = np.repeat(np.arange(n_levels), 2)
+    folds = _kfold_split_grouped(k=k, x=x)
+    groups_per_fold = [
+        len({g for g in np.unique(x) if folds[x == g][0] == fold}) for fold in range(1, k + 1)
+    ]
+    assert max(groups_per_fold) - min(groups_per_fold) <= 1
```
aloctavodia:
```suggestion
_, unique_indices = np.unique(x, return_index=True)
group_folds = folds[unique_indices]

_, fold_counts = np.unique(group_folds, return_counts=True)

assert fold_counts.max() - fold_counts.min() <= 1
```
Not tested, but I think we can use this instead

aloctavodia (approved):
LGTM!

tests/loo/test_helper_kfold.py
jordandeklerk: Added this and confirmed testing is good

PR #376, opened 2026-07-05
Replace absolufy-imports with ruff TID and rename ruff hook to ruff-check

PiyushPanwarFST:
- swap absolufy-imports for ruff's TID with ban-relative-imports

- rename the deprecated ruff hook id to ruff-check

- add pre-commit to dependabot so hook versions get bumped automatically

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -10,3 +10,13 @@ updates:
     actions:
       patterns:
         - "*"
+- package-ecosystem: "pre-commit"
+  directory: "/"
+  schedule:
+    interval: "monthly"
+  cooldown:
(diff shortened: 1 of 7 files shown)


PiyushPanwarFST:
Hii @OriolAbril  for `pre-commit` I used its own built-in method `pre-commit autoupdate --freeze`, which pins each hook to its SHA. 

One thing to note: the freeze method also updates the hooks to the latest version as part of the process, so pre-commit-hooks went from `v5.0.0 → v6.0.0`. From here, Dependabot takes care of the updates.

OriolAbril (approved):

PR #378, opened 2026-07-06
Add multi-chain uniformity test

aloctavodia:


src/arviz_stats/base/density.py
```diff
@@ -916,6 +916,65 @@ def _piet_c(self, ary):
...
+        for j in range(n_chains):
+            probs1 = hypergeom.cdf(ecdfs[:, j], samples, m, k)
+            probs2 = hypergeom.sf(ecdfs[:, j] - 1, samples, m, k)
+            gamma_chains[:, j] = 2 * np.minimum(probs1, probs2)
+
+        b_cauchy_vals = np.tan((0.5 - gamma_chains) * np.pi)
+        ps = 0.5 - np.arctan(b_cauchy_vals.mean(axis=1)) / np.pi
+
+        w_cauchy_vals = np.tan((0.5 - ps) * np.pi)
+        p_value = 0.5 - np.arctan(w_cauchy_vals.mean()) / np.pi
```
jordandeklerk, line 980: Haven't tested this much and maybe I'm misunderstanding, but for any discrete RV, isn't it the case that `cdf(x) + sf(x-1) = 1 + P(X = x) > 1`. So near the mode both terms exceed 0.5, so `2 * min(probs1, probs2)` can exceed 1. Also, `tan((0.5 - gamma) * pi)` has period `pi`, so as gamma crosses 1 the value jumps from large negative (evidence for the null) to large positive (evidence against it). 

I guess maybe just clipping would fix this?


```suggestion
        gamma_chains = np.empty((n_draws, n_chains))
        for j in range(n_chains):
            probs1 = hypergeom.cdf(ecdfs[:, j], samples, m, k)
            probs2 = hypergeom.sf(ecdfs[:, j] - 1, samples, m, k)
            gamma_chains[:, j] = np.minimum(2 * np.minimum(probs1, probs2), 1.0)

        b_cauchy_vals = np.tan((0.5 - gamma_chains) * np.pi)
        ps = 0.5 - np.arctan(b_cauchy_vals.mean(axis=1)) / np.pi

        w_cauchy_vals = np.tan((0.5 - ps) * np.pi)
        p_value = 0.5 - np.arctan(w_cauchy_vals.mean()) / np.pi
```

src/arviz_stats/base/density.py
```diff
@@ -916,6 +916,65 @@ def _piet_c(self, ary):
...
+            shape (n_draws,).
+        """
+        ary = np.asarray(ary)
+        if ary.ndim != 2:
+            raise ValueError("_mtc_c expects a 2D array")
+
+        n_chains, n_draws = ary.shape
+
+        if n_chains == 0 or n_draws == 0:
+            return np.nan, np.array([]), np.array([])
```
jordandeklerk: Shouldn't this test fail for only 1 chain as well?


```suggestion
        if n_draws == 0:
            return np.nan, np.array([]), np.array([])
        if n_chains < 2:
            raise ValueError("_mtc_c requires at least 2 chains for a multi-chain test.")
```

src/arviz_stats/base/density.py
```diff
@@ -916,6 +916,65 @@ def _piet_c(self, ary):
...
+        b_shapley_vals : ndarray
+            Between-chain Shapley contributions for each chain per draw,
+            shape (n_draws, n_chains).
+        w_shapley_vals : ndarray
+            Draw-wise Shapley contributions to the global p-value,
+            shape (n_draws,).
+        """
+        ary = np.asarray(ary)
+        if ary.ndim != 2:
+            raise ValueError("_mtc_c expects a 2D array")
```
jordandeklerk, line 944: Should we check for NaN/Inf here? I think a NaN/Inf is sorted to the end of each row by `np.sort`, so `np.searchsorted` miscounts the `ECDF`, and the function returns an incorrect p-value with no error. 

Worth testing a bit but maybe this is fine:


```suggestion
        ary = np.asarray(ary)
        if ary.ndim != 2:
            raise ValueError("_mtc_c expects a 2D array")
        if not np.all(np.isfinite(ary)):
            raise ValueError("_mtc_c requires finite fractional ranks (no NaN or Inf).")
```

tests/test_uniformity.py
```diff
@@ -273,3 +273,39 @@ def test_3d_batched(self, method):
         assert p_value.shape == (2, 3)
         assert shapley.shape == (2, 3, 80)
         assert shapley_unsorted.shape == (2, 3, 80)
+
+
+class TestMchainUniformity:
```
jordandeklerk, line 278: This only exercises the array layer. Should we be testing the dataarray and the accessor path here as well somewhere?

jordandeklerk (approved):
Left some comments for things that might be worth looking into. Otherwise LGTM!

src/arviz_stats/base/density.py
aloctavodia: You are right that this could lead to values larger than 1, but clipping can also bring problems. I tried to directly fix the problem.

PR #383, opened 2026-07-15
add docstub to arviz-stats

PiyushPanwarFST:
- add docstub config (docstub.toml, pre-commit hook, tox.ini)
- generate .pyi stubs for the package
- annotate the public API in the base folder (missing parameter types in docstrings)
- fix docstring type errors/typos so docstub passes on base
- ignore stub-only ruff rules (F401/F403/F811) for .pyi files


src/arviz_stats/base/array.py
```diff
@@ -15,7 +15,14 @@
 
 
 def process_chain_none(ary, chain_axis, draw_axis):
-    """Process array with chain and draw axis to cover the case ``chain_axis=None``."""
+    """Process array with chain and draw axis to cover the case ``chain_axis=None``.
+
+    Parameters
+    ----------
+    ary : array-like
+    chain_axis : int
```
OriolAbril, line 29:
```suggestion
    chain_axis : int or None
```

src/arviz_stats/base/array.py
```diff
@@ -24,7 +31,14 @@ def process_chain_none(ary, chain_axis, draw_axis):
 
 
 def process_chain_none_multi(*arys, chain_axis, draw_axis):
-    """Process multiple arrays with chain and draw axis to cover the case ``chain_axis=None``."""
+    """Process multiple arrays with chain and draw axis to cover the case ``chain_axis=None``.
+
+    Parameters
+    ----------
+    *arys : array-like
+    chain_axis : int
```
OriolAbril:
```suggestion
    chain_axis : int or None
```

src/arviz_stats/base/array.py
```diff
@@ -85,7 +99,7 @@ def hdi(
         circular : bool, default False
         max_modes : int, default 10
         skipna : bool, default False
-        **kwargs
+        **kwargs : dict, optional
```
OriolAbril:
```suggestion
        **kwargs
```

src/arviz_stats/base/array.py
```diff
@@ -850,7 +884,15 @@ def mchain_uniformity_test(self, ary, chain_axis=-2, draw_axis=-1, **kwargs):
         )
 
     def bayesian_r2(self, mu_pred, scale=None, scale_kind="sd", circular=False):
```
OriolAbril: would be better to include shape info as I don't think any arbitrary shape will work, I'll check to see what are the constraints

src/arviz_stats/base/array.py
```diff
@@ -431,7 +453,7 @@ def histogram(self, ary, bins=None, range=None, weights=None, axis=-1, density=T
         ----------
         ary : array-like
         bins : str, scalar or array-like, optional
```
OriolAbril: should be either

```
        bins : str or scalar or array-like, optional
```

or 

```
        bins : str | scalar or array-like, optional
```

src/arviz_stats/base/core.py
OriolAbril:
```suggestion
        axis : int or sequence of int or None, default None
```

OriolAbril:
```suggestion
    def _circular_standardize(self, ary: ArrayLike):
```

src/arviz_stats/base/core.py
```diff
@@ -233,7 +259,8 @@ def _get_bins(self, values, bins="arviz"):
 
         Returns
         -------
-        array with the bins
+        array-like
```
OriolAbril: ndarray

src/arviz_stats/base/core.py
OriolAbril:
```suggestion
        bins : int or str or array_like, default "arviz"
```

src/arviz_stats/base/dataarray.py
```diff
@@ -27,7 +27,16 @@ def __init__(self, array_class=None):
...
 
     def eti(self, da, prob=None, dim=None, method="linear", **kwargs):
-        """Compute eti on DataArray input."""
+        """Compute eti on DataArray input.
+
+        Parameters
+        ----------
+        da : DataArray
+        prob : float, optional
+        dim : sequence of hashable, optional
```
OriolAbril:
```suggestion
        dim : hashable or sequence of hashable, optional
```

src/arviz_stats/base/dataarray.py
```diff
@@ -113,7 +154,15 @@ def rhat(self, da, sample_dims=None, method="bulk"):
         )
 
     def rhat_nested(self, da, superchain_ids, sample_dims=None, method="rank"):
-        """Compute nested rhat on DataArray input."""
+        """Compute nested rhat on DataArray input.
+
+        Parameters
+        ----------
+        da : DataArray
+        superchain_ids : array-like
```
OriolAbril, line 183:
```suggestion
        superchain_ids : array-like of dtype int
```

docstub.toml
```diff
@@ -0,0 +1,41 @@
+[tool.docstub.type_prefixes]
+numbers = "numbers"
+xarray = "xarray"
+pandas = "pandas"
+numpyro = "numpyro"
+emcee = "emcee"
+cmdstanpy = "cmdstanpy"
+stan = "stan"
```
OriolAbril:
```suggestion
```

docstub.toml
```diff
@@ -0,0 +1,41 @@
...
+# TODO: define a meta/pseudo type for DataTree-like
+DataTree-like = "xarray.DataTree"
+DataArray = "xarray.DataArray"
+scalar = "numbers.Number"
+any = "Any"
+
+# arviz-stats specific
+InferenceData = "xarray.DataTree"
+idata-like = "xarray.DataTree"
+datatree = "xarray.DataTree"
```
OriolAbril: I would fix the capitalization of the places that use this in the type description

docstub.toml
```diff
@@ -0,0 +1,41 @@
...
+DataArrays = "xarray.DataArray"
```
OriolAbril: To match docstub syntax better, wherever this is used should be change to `DataArray` directly or `list of DataArray(s)` https://docstub.readthedocs.io/stable/typing_syntax.html#containers

docstub.toml
```diff
@@ -0,0 +1,41 @@
...
+DataFrame = "pandas.DataFrame"
+Namedtuple = "typing.NamedTuple"
```
OriolAbril: this should not be used, either full CamelCase or all lowercase

docstub.toml
```diff
@@ -0,0 +1,41 @@
...
+namedtuple = "typing.NamedTuple"
```
OriolAbril, line 34: This one will be allowed, we should also add an alias in `docs/source/conf.py` linking it to `` :term:`named tuple` ``

docstub.toml
```diff
@@ -0,0 +1,41 @@
...
+boolean = "bool"
+SamplingWrapper = "arviz_stats.loo.wrapper.SamplingWrapper"
```
OriolAbril:
```suggestion
SamplingWrapper = "arviz_stats.SamplingWrapper"
```

PiyushPanwarFST: Hey @OriolAbril for SamplingWrapper I used `arviz_stats.loo.SamplingWrapper` instead of the top-level `arviz_stats.SamplingWrapper`, because the top-level path doesn't resolve in docstub because it's imported inside a try block in `__init__.py`

OriolAbril: sounds good, not reviewing again for now, do let me know if you have any doubts and I'll review again

PiyushPanwarFST: Yaa sure

src/arviz_stats/base/array.py
```diff
@@ -299,7 +314,14 @@ def psislw(self, ary, r_eff=1, axis=-1):
         return psl_ufunc(ary, out_shape=[(ary.shape[i] for i in axes), []], r_eff=r_eff)
 
     def bfmi(self, ary, chain_axis=-2, draw_axis=-1):
-        """Calculate the estimated Bayesian fraction of missing information."""
+        """Calculate the estimated Bayesian fraction of missing information.
+
+        Parameters
+        ----------
+        ary : array-like
+        chain_axis : int, default -2
```
OriolAbril:
```suggestion
        chain_axis : int or None, default -2
```

src/arviz_stats/base/array.py
OriolAbril:
```suggestion
        chain_axis : int or None, default -2
```

OriolAbril (commented):
left a couple small comments, in general the main missing thing are "Returns" sections is some of the docstrings. I'll review in depth once you push the rest of the docstring updates

src/arviz_stats/base/array.py
```diff
@@ -15,7 +15,14 @@
...
 def process_chain_none(ary, chain_axis, draw_axis):
-    """Process array with chain and draw axis to cover the case ``chain_axis=None``."""
+    """Process array with chain and draw axis to cover the case ``chain_axis=None``.
+
+    Parameters
+    ----------
+    ary : array-like
+    chain_axis : int or None
+    draw_axis : int
+    """
```
OriolAbril, line 31: add returns section

src/arviz_stats/base/array.py
```diff
@@ -24,7 +31,14 @@ def process_chain_none(ary, chain_axis, draw_axis):
...
 def process_chain_none_multi(*arys, chain_axis, draw_axis):
-    """Process multiple arrays with chain and draw axis to cover the case ``chain_axis=None``."""
+    """Process multiple arrays with chain and draw axis to cover the case ``chain_axis=None``.
+
+    Parameters
+    ----------
+    *arys : array-like
+    chain_axis : int or None
+    draw_axis : int
+    """
```
OriolAbril, line 52: add returns section

src/arviz_stats/base/array.py
OriolAbril, line 66: add returns section

OriolAbril, line 121: add returns section

src/arviz_stats/base/array.py
```diff
@@ -151,11 +165,11 @@ def ess(self, ary, chain_axis=-2, draw_axis=-1, method="bulk", relative=False, p
...
-        chain_axis : int, default -2
+        chain_axis : int or None, default -2
         draw_axis : int, default -1
         method : str, default "bulk"
         relative : bool, default False
-        prob : float or (float, float), default None
+        prob : float or tuple of (float, float), default None
             When using the array interface, `prob` is a required argument for
             the "tail", "quantile" (as float) and "local" (as tuple) methods.
         """
```
OriolAbril, line 206: add returns section

src/arviz_stats/base/array.py
```diff
@@ -186,7 +200,7 @@ def rhat(self, ary, chain_axis=-2, draw_axis=-1, method="rank"):
         Parameters
         ----------
         ary : array-like
-        chain_axis : int, default -2
+        chain_axis : int or None, default -2
         draw_axis : int, default -1
         method : str, default "rank"
         """
```
OriolAbril, line 241: add returns section

OriolAbril (commented):
I have commented on several functions, these or some others you might have had to fix already while adding mypy but the comment is not specific to the functions with comments and it is general to the whole repo

src/arviz_stats/visualization.pyi
```diff
@@ -0,0 +1,97 @@
...
+    group: Hashable = ...,
+    var_names: str | list[str] | None = ...,
+    filter_vars: Literal[None, "like", "regex"] | None = ...,
+    coords: dict | None = ...,
+    method: str = ...,
+    circular: bool = ...,
+    max_modes: int = ...,
+    skipna: bool = ...,
+    **kwargs: Any,
+) -> NDArray: ...
```
OriolAbril: These return types need fixing, see https://github.com/arviz-devs/arviz-stats/blob/96eac2b3df5b2b2037e32981b30cbfbfce31a672/src/arviz_stats/visualization.py#L70, they use commas instead of `or` or `|` so only the first type mentioned is present instead of all possible input types

src/arviz_stats/visualization.pyi
```diff
@@ -0,0 +1,97 @@
...
+from _typeshed import Incomplete
+from arviz_base import convert_to_dataset
+from numpy.typing import ArrayLike, NDArray
+from xarray import DataArray, Dataset
+
+from arviz_stats.utils import _apply_multi_input_function, get_function
+from arviz_stats.validate import validate_ci_prob, validate_dims
+
+def hdi(
+    data: ArrayLike,
```
OriolAbril, line 17: same thing is happening here, docstring uses commas instead of `or` or `|`

src/arviz_stats/visualization.pyi
```diff
@@ -0,0 +1,97 @@
...
+    **kwargs: Any,
+) -> NDArray: ...
+def histogram(
+    data: ArrayLike,
+    dim: Sequence[Hashable] | None = ...,
+    group: Hashable = ...,
+    var_names: str | list[str] | None = ...,
+    filter_vars: Literal[None, "like", "regex"] | None = ...,
+    coords: dict | None = ...,
+    bins: Incomplete = ...,
```
OriolAbril: the docstring for histogram has a typo in the `bins` argument and uses `bind` instead. There is also a docstring entry for `**kwargs` which is not present in histogram and should also be removed from the docstring

src/arviz_stats/base/array.py
```diff
@@ -15,7 +15,20 @@
...
+
+    Parameters
+    ----------
+    ary : array-like
+    chain_axis : int or None
+    draw_axis : int
+
+    Returns
+    -------
+    ary : array-like
```
OriolAbril, line 28: Here we are not doing anything to `ary` so if it is not an ndarray it will keep its type and it is ok to use `array-like` in the output type, but as we have discussed, for outputs this is the exception and not the norm

src/arviz_stats/base/array.py
```diff
@@ -24,7 +37,19 @@ def process_chain_none(ary, chain_axis, draw_axis):
 
 
 def process_chain_none_multi(*arys, chain_axis, draw_axis):
```
OriolAbril, line 39:
```suggestion
def process_chain_none_multi(*arys, chain_axis, draw_axis) -> tuple[*tuple[array-like, ...], int, int]:
```

docstub doesn't allow for this [very specific syntax](https://typing.python.org/en/latest/spec/tuples.html#unpacked-tuple-form) so in this case it is better to use the more accurate inline type hint (which will take preference over any docstring if present). This is the same we do for example for `visuals` or `aes_by_visuals` arguments in arviz-plots

src/arviz_stats/base/array.py
```diff
@@ -39,6 +64,11 @@ def process_ary_axes(ary, axes):
     ----------
     ary : array_like
     axes : int or sequence of int
+
+    Returns
+    -------
+    ary : array-like
+    axes : ndarray
```
OriolAbril, line 71:
```suggestion
    axes : ndarray of dtype int and shape (N,)
```

we know dtype and shape of output so we can add it to the docstring

src/arviz_stats/base/array.py
```diff
@@ -89,6 +119,10 @@ def hdi(
             Only used for multimodal methods with continuous data.
             Passed to kde computation with a ``bw`` default of "taylor" for
             circular data, "isj" otherwise.
+
+        Returns
+        -------
+        array-like
```
OriolAbril, line 125:
```suggestion
        ndarray
```

The output here will always be an ndarray, we generate a numpy ufunc with `make_ufunc` which generates an empty array and fills it as needed taking into account the input dimensions and dimensions to reduce. Except the helpers above that only check the shape of the inputs and usually leave the input as is, everything else under here returns a whole new object which is always an ndarray

src/arviz_stats/base/array.py
```diff
@@ -39,6 +64,11 @@ def process_ary_axes(ary, axes):
     ----------
     ary : array_like
     axes : int or sequence of int
+
+    Returns
+    -------
+    ary : array-like
```
OriolAbril, line 70: Here we call `np.transpose` on `ary` which will generally cast the output to ndarray but there are some edgecases like `DataArray` where if it is the input type it will be preserved in the output so it can be argued using `array-like` is also correct here

src/arviz_stats/base/array.py
```diff
@@ -706,7 +785,23 @@ def qds(
...
-        """Compute quantile dot."""
+        """Compute quantile dot.
+
+        Parameters
+        ----------
+        ary : array-like
+        nquantiles : int, default 100
+        binwidth : float, optional
+        dotsize : float, default 1
+        stackratio : float, default 1
```
OriolAbril, line 803: for these values we don't care if they are float or integer (in fact, the default for dotsize and stackratio is already an int which breaks the type hint) so it should use `scalar` instead

src/arviz_stats/base/core.py
```diff
@@ -17,12 +18,46 @@
...
     def fft(self, x):
+        """Compute the one-dimensional discrete Fourier Transform.
+
+        Parameters
+        ----------
+        x : array-like
+
+        Returns
+        -------
+        array-like
```
OriolAbril, line 29: also ndarray for most functions in this file as output

src/arviz_stats/base/core.py
```diff
@@ -82,10 +125,18 @@ def circular_mean(self, ary):
...
+        ary : array-like
+
+        Returns
+        -------
+        float
         """
         return circmean(ary, high=np.pi, low=-np.pi)
 
-    def _circular_standardize(self, ary):
+    def _circular_standardize(self, ary: ArrayLike):
```
OriolAbril, line 139:
```suggestion
    def _circular_standardize(self, ary: ArrayLike) -> NDArray[float]:
```

We know the dtype but not the shape so we can use the NDArray helper from numpy: https://numpy.org/devdocs/reference/typing.html#numpy.typing.NDArray. As we have discussed for functions like this it is ok to not have the full docstring, but the output type should still be present

src/arviz_stats/base/core.py
```diff
@@ -149,7 +208,7 @@ def eti(self, ary, prob, axis=-1, method="linear", skipna=False, weights=None):
             weights=weights,
         )
 
-    def _float_rankdata(self, ary):
+    def _float_rankdata(self, ary: ArrayLike):
```
OriolAbril, line 211:
```suggestion
    def _float_rankdata(self, ary: ArrayLike) -> np.ndarray[tuple[int], np.dtype[int]]:
```

src/arviz_stats/base/core.py
```diff
@@ -163,7 +222,7 @@ def _float_rankdata(self, ary):
         ranks[np.argsort(ary, axis=None)] = np.arange(1, ary.size + 1)
         return ranks
 
-    def _compute_ranks(self, ary, relative=False):
+    def _compute_ranks(self, ary: ArrayLike, relative: bool = False):
```
OriolAbril, line 225:
```suggestion
    def _compute_ranks(self, ary: ArrayLike, relative: bool = False) -> np.ndarray:
```

src/arviz_stats/base/core.py
```diff
@@ -264,12 +324,19 @@ def _get_bins(self, values, bins="arviz"):
...
     # pylint: disable=redefined-builtin
-    def _histogram(self, ary, bins=None, range=None, weights=None, density=True):
+    def _histogram(
+        self,
+        ary: ArrayLike,
+        bins: int | str | ArrayLike | None = None,
+        range: tuple | None = None,
+        weights: ArrayLike | None = None,
+        density: bool = True,
+    ):
```
OriolAbril, line 334:
```suggestion
    ) -> tuple[np.ndarray, np.ndarray]:
```

src/arviz_stats/base/core.py
```diff
@@ -390,7 +470,14 @@ def _hdi_from_bin_probabilities(self, bins, bin_probs, prob, circular, dx):
 
         return self._interval_points_to_bounds(intervals, probs_in_interval, dx, circular)
 
-    def _interval_points_to_bounds(self, points, probs, dx, circular, period=2 * np.pi):
+    def _interval_points_to_bounds(
+        self,
+        points: ArrayLike,
+        probs: ArrayLike,
```
OriolAbril, line 476:
```suggestion
        points: np.ndarray,
        probs: np.ndarray,
```

The first line uses `probs.cumsum()` and later on `points` is indexed with arrays so most array-like things like tuples or lists are not valid inputs.

src/arviz_stats/base/core.py
```diff
@@ -413,7 +500,9 @@ def _interval_points_to_bounds(self, points, probs, dx, circular, period=2 * np.
 
         return interval_bounds, interval_probs
 
-    def _pad_hdi_to_maxmodes(self, hdi_intervals, interval_probs, max_modes):
+    def _pad_hdi_to_maxmodes(
+        self, hdi_intervals: ArrayLike, interval_probs: ArrayLike, max_modes: int
```
OriolAbril, line 504:
```suggestion
        self, hdi_intervals: np.ndarray, interval_probs: ArrayLike, max_modes: int
```

src/arviz_stats/base/core.py
```diff
@@ -451,22 +540,22 @@ def _median(self, ary, round_to=None, skipna=False, axis=None):
...
-        axis : int, sequence of int or None, default None
+        axis : int or sequence of int or None, default None
             Axis or axes along which to compute the median.
         """
         if skipna:
             ary = ary[~np.isnan(ary)]
         return round_num(np.median(ary, axis=axis), round_to)
 
-    def _mode(self, ary, round_to=None, skipna=False):
+    def _mode(self, ary: ArrayLike, round_to: int | str | None = None, skipna: bool = False):
```
OriolAbril, line 558:
```suggestion
    def _mode(self, ary: np.ndarray, round_to: int | str | None = None, skipna: bool = False) -> np.ndarray:
```

src/arviz_stats/base/density.py
```diff
@@ -94,14 +97,38 @@ def _root(self, function, n, args, x, grid_range):
...
 
     def bw_scott(self, x, x_std=None, **kwargs):  # pylint: disable=unused-argument, no-self-use
-        """Scott's Rule."""
+        """Scott's Rule.
+
+        Parameters
+        ----------
+        x : array-like
+        x_std : float, optional
+        **kwargs : dict, optional
```
OriolAbril, line 106:
```suggestion
        **kwargs
```

src/arviz_stats/base/density.py
```diff
@@ -94,14 +97,38 @@ def _root(self, function, n, args, x, grid_range):
...
 
     def bw_silverman(self, x, x_std=None, **kwargs):  # pylint: disable=unused-argument
-        """Silverman's Rule."""
+        """Silverman's Rule.
+
+        Parameters
+        ----------
+        x : array-like
+        x_std : float, optional
+        **kwargs : dict, optional
```
OriolAbril, line 125:
```suggestion
        **kwargs
```

src/arviz_stats/base/density.py
```diff
@@ -169,16 +221,19 @@ def get_bw(self, x, bw, grid_counts=None, x_std=None, grid_range=None):
 
         Parameters
         ----------
-        x : 1-D numpy array
+        x : array-like
```
OriolAbril, line 224:
```suggestion
        x : array-like of shape (n,)
```

src/arviz_stats/accessors.py
```diff
@@ -524,13 +797,25 @@ def _apply(self, func, **kwargs):
         return out_dt
 
     def ecdf(self, *args, **kwargs):
```
PiyushPanwarFST: Hey Oriol, `Dt.ecdf` had no Returns section so docstub typed it -> None. But the parent `_BaseAccessor.ecdf` returns Dataset, so Mypy rejects the override.

  Screenshot 2026-08-17 at 4 46 03 PM

   So to fix this I added Returns / NoReturn in the docstring let me know if some better way exist ?

src/arviz_stats/summary.py
```diff
@@ -372,7 +371,7 @@ def _build_fmt_map(summary_result, round_val):
     return fmt_map
 
 
-class SummaryDataFrame(pd.DataFrame):
+class SummaryDataFrame(pd.DataFrame):  # type: ignore[misc]
```
PiyushPanwarFST: Hey Oriol, `SummaryDataFrame` is built on top of `pd.DataFrame`, but pandas has no type info in its stubs, so Mypy treats DataFrame as Any and it won't let us build a class on top of Any.

  Screenshot 2026-08-17 at 4 34 42 PM

  so to fix it I added # type: ignore[misc] on the class line.

src/arviz_stats/base/diagnostics.pyi
```diff
@@ -0,0 +1,212 @@
...
+    def _ess_folded(self, ary: ArrayLike, relative: bool = ...) -> None: ...
+    def _ess_median(self, ary: ArrayLike, relative: bool = ...) -> None: ...
+    def _ess_mad(self, ary: ArrayLike, relative: bool = ...) -> None: ...
+    def _ess_identity(self, ary: ArrayLike, relative: bool = ...) -> None: ...
+    def _mcse_mean(self, ary: ArrayLike, circular: bool = ...) -> None: ...
+    def _mcse_sd(self, ary: ArrayLike) -> None: ...
+    def _mcse_median(self, ary: ArrayLike, circular: bool = ...) -> None: ...
+    def _mcse_quantile(
+        self, ary: ArrayLike, prob: float, circular: bool = ...
+    ) -> None: ...
```
OriolAbril, line 50: again, it's ok to use inline type hints in the `.py` file for all these but none of these functions returns `None`, they all return `ndarray`, the inline type hints also need a return type

src/arviz_stats/loo/loo_kfold_helper.py
```diff
@@ -24,26 +27,34 @@
...
-        "sample_dims",
-        "obs_dims",
-        "n_data_points",
-        "n_samples",
-        "folds",
-        "k",
-    ],
-)
 
+class KfoldInputs(NamedTuple):
```
PiyushPanwarFST: Hey @OriolAbril here in `KfoldInputs = namedtuple(...) ` is a variable using it as a type annotation in `_compute_kfold_results `on line 115 causes Mypy [valid-type].

  Screenshot 2026-08-17 at 4 22 40 PM

  So I converted  it to `class KfoldInputs(NamedTuple)` following the same pattern as SplitMomentMatch / LooMomentMatchResult in loo_moment_match.py. let me know if any adjustment needed ?

src/arviz_stats/base/__init__.py
```diff
@@ -10,3 +10,5 @@
     from arviz_stats.base.dataarray import dataarray_stats
 except ModuleNotFoundError:
     pass
+
+__all__ = ["array_stats", "dataarray_stats"]
```
PiyushPanwarFST: Hey @OriolAbril,  `metrics.py` does from `arviz_stats.base import array_stats`, but `base/__init__.py` imports it internally and doesn't mark it as public, so Mypy flagged it.

  Screenshot 2026-08-17 at 5 01 57 PM

  To fix it I added `__all__ = ["array_stats", "dataarray_stats"]` in `base/__init__.py` (same as what loo/__init__.py already does). Let me know if that looks right.

src/arviz_stats/visualization.py
OriolAbril, line 31: I commented on this on the pyi file, this is using commas so the stub only lists array-like instead of all the types

.pre-commit-config.yaml
OriolAbril, line 1: I see the commit message talks about fixing mypy errors but there is no mypy being added here

src/arviz_stats/numba/array.py
```diff
@@ -168,9 +191,11 @@ def kde(self, ary, axis=-1, circular=False, grid_len=512, **kwargs):
         return self.kde_ufunc(ary, np.empty(grid_len), circular, bw, adaptive)
```
PiyushPanwarFST: Hey @OriolAbril here Mypy raising 2 errors , as you can see in .pyi file the module-level.

`NumbaArray.histogram.__doc__ = ...` in `numba/array.py` makes docstub add 3 extra lines to the stub  (`NumbaArray: Incomplete, histogram: Incomplete, __doc__: Incomplete`) also there is an conflict with the already-defined `NumbaArray` class and `__doc__. `

Screenshot 2026-08-17 at 1 06 58 PM

Screenshot 2026-08-17 at 1 02 57 PM

 So to resolve this error I wrapped it in a small private function `_extend_histogram_docstring()`.

src/arviz_stats/base/dataarray.py
PiyushPanwarFST: Hey @OriolAbril , I tried adding `array-like of dtype int` as you suggested for `superchain_ids`,  but Mypy flagged it and gave me this error.

Screenshot 2026-08-17 at 12 54 17 PM

So resolve it I reverted back the change and kept again `array-like`.

.pre-commit-config.yaml
PiyushPanwarFST: Actually till now I have been running mypy locally to fix the errors and hadn't wired the hook yet, but I will fix this and made changes according to all other suggestion and will raise a new commit soon.

src/arviz_stats/base/array.py
PiyushPanwarFST: Here I kept the type as ndarray with the dtype/shape as a description line.

```
    axes : ndarray
        Integer dtype and shape ``(N,)``
```

Because writing it inline makes docstub emit `NDArray[int]`, which mypy rejects

Screenshot 2026-08-19 at 3 42 28 PM

src/arviz_stats/base/core.py
```diff
@@ -136,7 +136,7 @@ def circular_mean(self, ary):
         """
         return circmean(ary, high=np.pi, low=-np.pi)
 
-    def _circular_standardize(self, ary: ArrayLike):
+    def _circular_standardize(self, ary: ArrayLike) -> NDArray[np.floating]:
```
PiyushPanwarFST: Here for` _circular_standardize` I used `-> NDArray[np.floating] `return type instead of `-> NDArray[float]` because mypy rejects the plain float.

Screenshot 2026-08-19 at 3 27 58 PM

src/arviz_stats/base/core.py
```diff
@@ -208,7 +208,7 @@ def eti(self, ary, prob, axis=-1, method="linear", skipna=False, weights=None):
             weights=weights,
         )
 
-    def _float_rankdata(self, ary: ArrayLike):
+    def _float_rankdata(self, ary: ArrayLike) -> NDArray[np.integer]:
```
PiyushPanwarFST: For `_float_rankdata` the plain int in the dtype makes mypy error: `(np.ndarray[tuple[int], np.dtype[int]])` 

Screenshot 2026-08-19 at 3 34 21 PM

So I used `NDArray[np.integer]` gives the int dtype and passes.

src/arviz_stats/base/core.py
```diff
@@ -550,12 +554,22 @@ def _median(self, ary, round_to=None, skipna=False, axis=None):
...
+
+        Returns
+        -------
+        float
         """
         if skipna:
             ary = ary[~np.isnan(ary)]
         return round_num(np.median(ary, axis=axis), round_to)
 
-    def _mode(self, ary: ArrayLike, round_to: int | str | None = None, skipna: bool = False):
```
PiyushPanwarFST: - Applied the `ary: np.ndarray` input as you suggested.
-  But I kept the output as scalar instead of `-> np.ndarray` because `_mode` returns one value on every path: `np.nan`, `ary.item()`, `round_num(...)`and `vals[cnts.argmax()]`. 
- So scalar matches what it actually returns. Let me know if this is correct ?

PR #386, opened 2026-07-19
Add moment matching arg to main `loo()` function

jordandeklerk:
Adds a moment macthing arg to the main `loo()` function

### DIFF
--- a/src/arviz_stats/loo/loo.py
+++ b/src/arviz_stats/loo/loo.py
@@ -16,6 +16,7 @@ from arviz_stats.loo.loo_helper import (
     _warn_pareto_k,
     _warn_pointwise_loo,
 )
+from arviz_stats.loo.loo_moment_match import loo_moment_match
 from arviz_stats.utils import ELPDData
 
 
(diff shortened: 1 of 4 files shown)


jordandeklerk:
Reorganized this a bit to just allow automatic moment matching through the main `loo()` function. If users want to use custom functions and change the defaults for the other moment matching parameters, that would happen through the `loo_moment_match()` function. 

I think this is cleaner IMO instead of having a large number of parameters for the main `loo()` function.

Ready for review and thoughts on this @aloctavodia 

---
Note: Needed to handle arrays a little differently in the pymc/bmbi helper file for moment matching because of a numba error with arrays of somesort. Those changes are for that reason.

src/arviz_stats/loo/loo.py
```diff
@@ -73,7 +76,18 @@ def loo(
...
+        If True, apply the moment matching algorithm described in [1]_ to observations
+        with high Pareto k values. Requires ``model``. This is equivalent to computing
+        LOO with ``pointwise=True`` and passing the result to :func:`loo_moment_match`.
+        Cannot be combined with ``mixture=True``, ``log_lik_fn`` or ``log_jacobian``.
+        For other models, or for fine-grained control over the moment matching algorithm,
+        use :func:`loo_moment_match` directly.
+    model : Model, optional
+        Required when ``moment_match=True``. A model object used to automatically build
+        the quantities needed for moment matching. Currently supported models are PyMC
+        and Bambi.
```
aloctavodia:
```suggestion
        the quantities needed for moment matching. Currently, only models from PyMC
        and Bambi are supported.
```

aloctavodia:
I agree with the `loo`, `loo_moment_match()` matching separation.

PR #389, opened 2026-07-21
Fix `bayesian_r2` silently using Bernoulli pseudo-variance when scale is omitted.

### LINKED ISSUE #388: `bayesian_r2` silently uses the Bernoulli pseudo-variance when `scale` is omitted, returning R² outside [0, 1] for continuous models
When `bayesian_r2` is called without a `scale` argument, it silently assumes a Bernoulli/Tjur-style model and computes the residual (pseudo-)variance as `mean(mu_pred * (1 - mu_pred))`. That expression is only meaningful when `mu_pred` are probabilities in [0, 1]. For a continuous/Gaussian model whose predicted means are real-valued, `mu_pred * (1 - mu_pred` can be negative, so the denominator `var_fit + scale` can approach zero or go negative, and the function returns R² values well outside [0, 1] (including large positive and negative spikes) with no warning and no error.

The non-negativity check that would catch this (`if np.any(scale < 0): raise ValueError("Scale must be non-negative."`)) lives only in the branch that handles a user-supplied scale; the internally-computed Bernoulli pseudo-variance bypasses it entirely. So a user who simply forgets scale (easy, since it's optional) gets a plausible-looking but invalid number rather than a failure.

This bit me in production: a Gaussian regression's per-period "R²" was fed to a dashboard and showed values from roughly −5 to +13, which is what led us here.

Minimal reproducible example

```
import numpy as np
import arviz as az
from arviz_stats import bayesian_r2

rng = np.random.default_rng(0)
n_obs, n_draws = 50, 500

# Gaussian regression: predicted means are real-valued, NOT probabilities in [0, 1].
mu_true = rng.normal(0.0, 0.05, size=n_obs)
pred_mean = mu_true[None, None, :] + rng.normal(0.0, 0.02, size=(1, n_draws, n_obs))

idata = az.from_dict(
    {"posterior": {"mu": pred_mean}},
    coords={"obs": np.arange(n_obs)},
    dims={"mu": ["obs"]},
)

# 'scale' (the modelled residual sd/variance) omitted -> silent Bernoulli pseudo-variance fallback
r2 = bayesian_r2(idata, pred_mean="mu", summary=False)
print("R2 min = %.3f, max = %.3f" % (r2.min(), r2.max()))
print("fraction outside [0, 1] = %.2f" % np.mean((r2 < 0) | (r2 > 1)))
```

Actual output

``R2 min = -10.925, max = 41.988
fraction outside [0, 1] = 0.07``

No warning or error is emitted.

CamSalda:
Closes https://github.com/arviz-devs/arviz-stats/issues/388 

## Changes

- `_bayesian_r2` now emits a `UserWarning` whenever the Bernoulli pseudo-variance fallback is
  used (i.e. `scale` is not provided), pointing the user to pass the modelled residual `scale`
  for continuous models.
- The computed R² is range-checked; if any value falls outside [0, 1] a `ValueError` is raised.
  With a valid `scale >= 0` the ratio is mathematically always in [0, 1], so this only fires on
  the misused Bernoulli fallback and does not affect correct callers.
- Added two tests in `tests/test_metrics.py`: one asserting the warning (and result in [0, 1])
  for probability-valued `mu_pred`, one asserting the error for the continuous reproducer from
  the issue.

After this PR the same call raises `ValueError` `(scale omitted for a continuous model)`, and a
genuine Bernoulli use (probabilities in [0, 1], scale omitted) warns instead of failing.

### DIFF
--- a/src/arviz_stats/base/diagnostics.py
+++ b/src/arviz_stats/base/diagnostics.py
@@ -1537,6 +1537,13 @@ class _DiagnosticsBase(_CoreBase):
             if circular:
                 raise ValueError("scale must be provided for circular response.")
 
+            warnings.warn(
+                "`scale` not provided; assuming a Bernoulli-like model and using the Tjur "
+                "pseudo-variance mean(mu_pred * (1 - mu_pred)). For continuous models pass the "
+                "modelled residual `scale`.",
+                UserWarning,
+                stacklevel=2,
+            )
             # Bernoulli-like models: use Tjur’s pseudo-variance
             scale = np.mean(mu_pred * (1 - mu_pred), axis=1)
         else:
@@ -1555,7 +1562,14 @@ class _DiagnosticsBase(_CoreBase):
             return 1 - scale
 
         var_y_est = np.var(mu_pred, axis=1, ddof=1)
-        return var_y_est / (var_y_est + scale)
+        r2 = var_y_est / (var_y_est + scale)
+        if np.any((r2 < 0) | (r2 > 1)):
+            raise ValueError(
+                "Computed R² outside [0, 1]. This usually means the Bernoulli "
+                "pseudo-variance fallback was applied to a non-probability `mu_pred`; "
+                "pass the modelled residual `scale` for continuous models."
+            )
+        return r2
 
     @staticmethod
     def _residual_r2(y_obs, mu_pred, circular):
--- a/src/arviz_stats/base/diagnostics.py
+++ b/src/arviz_stats/base/diagnostics.py
@@ -1566,9 +1566,10 @@ class _DiagnosticsBase(_CoreBase):
         r2 = var_y_est / (var_y_est + scale)
         if np.any((r2 < 0) | (r2 > 1)):
             raise ValueError(
-                "Computed R² outside [0, 1]. This usually means the Bernoulli "
-                "pseudo-variance fallback was applied to a non-probability `mu_pred`; "
-                "pass the modelled residual `scale` for continuous models."
+                "Computed R-squared outside [0, 1]. The Bernoulli pseudo-variance "
+                "`mean(mu_pred * (1 - mu_pred))` was used because `scale` was not provided, "
+                "but `mu_pred` are not probabilities in [0, 1]. Pass the modelled residual "
+                "`scale` for continuous models."
             )
         return r2
 
--- a/src/arviz_stats/base/diagnostics.py
+++ b/src/arviz_stats/base/diagnostics.py
@@ -1538,9 +1538,10 @@ class _DiagnosticsBase(_CoreBase):
                 raise ValueError("scale must be provided for circular response.")
 
             warnings.warn(
-                "`scale` not provided; assuming a Bernoulli-like model and using the Tjur "
-                "pseudo-variance mean(mu_pred * (1 - mu_pred)). For continuous models pass the "
-                "modelled residual `scale`.",
+                "`scale` was not provided, so a Bernoulli/binary model is assumed and the "
+                "pseudo-variance `mean(mu_pred * (1 - mu_pred))` is used as the residual "
+                "variance. This is only valid when `mu_pred` are probabilities in [0, 1]. "
+                "For continuous (e.g. Gaussian) models, pass the modelled residual `scale`.",
                 UserWarning,
                 stacklevel=2,
             )
--- a/tests/test_metrics.py
+++ b/tests/test_metrics.py
@@ -100,7 +100,6 @@ def test_residual_r2_invalid_shapes():
 
 
 def test_bayesian_r2_no_scale_warns():
-    # Bernoulli-appropriate mu_pred in [0, 1], scale omitted -> Tjur fallback + warning.
     rng = np.random.default_rng(0)
     mu_pred = rng.uniform(0.1, 0.9, size=(200, 40))
     with pytest.warns(UserWarning, match="Bernoulli"):
--- a/tests/test_metrics.py
+++ b/tests/test_metrics.py
@@ -107,6 +107,7 @@ def test_bayesian_r2_no_scale_warns():
     assert np.all((r2 >= 0) & (r2 <= 1))
 
 
+@pytest.mark.filterwarnings("ignore::UserWarning")
 def test_bayesian_r2_continuous_no_scale_errors():
     rng = np.random.default_rng(0)
     mu_pred = rng.normal(0.0, 0.05, size=(200, 40))
--- a/tests/test_metrics.py
+++ b/tests/test_metrics.py
@@ -108,8 +108,6 @@ def test_bayesian_r2_no_scale_warns():
 
 
 def test_bayesian_r2_continuous_no_scale_errors():
-    # Continuous mu_pred (real-valued), scale omitted -> Bernoulli fallback yields R² outside
-    # [0, 1], which must raise instead of returning silently.
     rng = np.random.default_rng(0)
     mu_pred = rng.normal(0.0, 0.05, size=(200, 40))
     with pytest.raises(ValueError, match=r"outside \[0, 1\]"):
--- a/tests/test_metrics.py
+++ b/tests/test_metrics.py
@@ -99,6 +99,24 @@ def test_residual_r2_invalid_shapes():
         array_stats.residual_r2(y_true, y_pred)
 
 
+def test_bayesian_r2_no_scale_warns():
+    # Bernoulli-appropriate mu_pred in [0, 1], scale omitted -> Tjur fallback + warning.
+    rng = np.random.default_rng(0)
+    mu_pred = rng.uniform(0.1, 0.9, size=(200, 40))
+    with pytest.warns(UserWarning, match="Bernoulli"):
+        r2 = array_stats.bayesian_r2(mu_pred)
+    assert np.all((r2 >= 0) & (r2 <= 1))
+
+
+def test_bayesian_r2_continuous_no_scale_errors():
+    # Continuous mu_pred (real-valued), scale omitted -> Bernoulli fallback yields R² outside
+    # [0, 1], which must raise instead of returning silently.
+    rng = np.random.default_rng(0)
+    mu_pred = rng.normal(0.0, 0.05, size=(200, 40))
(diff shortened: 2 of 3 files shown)


tests/test_metrics.py
```diff
@@ -99,6 +99,24 @@ def test_residual_r2_invalid_shapes():
         array_stats.residual_r2(y_true, y_pred)
 
 
+def test_bayesian_r2_no_scale_warns():
+    # Bernoulli-appropriate mu_pred in [0, 1], scale omitted -> Tjur fallback + warning.
+    rng = np.random.default_rng(0)
+    mu_pred = rng.uniform(0.1, 0.9, size=(200, 40))
+    with pytest.warns(UserWarning, match="Bernoulli"):
+        r2 = array_stats.bayesian_r2(mu_pred)
+    assert np.all((r2 >= 0) & (r2 <= 1))
```
jordandeklerk, line 107:
```suggestion
def test_bayesian_r2_no_scale_warns():
    rng = np.random.default_rng(0)
    mu_pred = rng.uniform(0.1, 0.9, size=(200, 40))
    with pytest.warns(UserWarning, match="Bernoulli"):
        r2 = array_stats.bayesian_r2(mu_pred)
    assert np.all((r2 >= 0) & (r2 <= 1))
```

tests/test_metrics.py
```diff
@@ -99,6 +99,24 @@ def test_residual_r2_invalid_shapes():
...
+
+
+def test_bayesian_r2_continuous_no_scale_errors():
+    # Continuous mu_pred (real-valued), scale omitted -> Bernoulli fallback yields R² outside
+    # [0, 1], which must raise instead of returning silently.
+    rng = np.random.default_rng(0)
+    mu_pred = rng.normal(0.0, 0.05, size=(200, 40))
+    with pytest.raises(ValueError, match=r"outside \[0, 1\]"):
+        array_stats.bayesian_r2(mu_pred)
```
jordandeklerk, line 115:
```suggestion
def test_bayesian_r2_continuous_no_scale_errors():
    rng = np.random.default_rng(0)
    mu_pred = rng.normal(0.0, 0.05, size=(200, 40))
    with pytest.raises(ValueError, match=r"outside \[0, 1\]"):
        array_stats.bayesian_r2(mu_pred)
```

src/arviz_stats/base/diagnostics.py
```diff
@@ -1537,6 +1537,13 @@ def _bayesian_r2(mu_pred, scale, scale_kind, circular):
             if circular:
                 raise ValueError("scale must be provided for circular response.")
 
+            warnings.warn(
+                "`scale` not provided; assuming a Bernoulli-like model and using the Tjur "
+                "pseudo-variance mean(mu_pred * (1 - mu_pred)). For continuous models pass the "
+                "modelled residual `scale`.",
+                UserWarning,
+                stacklevel=2,
+            )
```
jordandeklerk, line 1547:
```suggestion
            warnings.warn(
                "`scale` was not provided, so a Bernoulli/binary model is assumed and the "
                "pseudo-variance `mean(mu_pred * (1 - mu_pred))` is used as the residual "
                "variance. This is only valid when `mu_pred` are probabilities in [0, 1]. "
                "For continuous (e.g. Gaussian) models, pass the modelled residual `scale`.",
                UserWarning,
                stacklevel=2,
            )
```

src/arviz_stats/base/diagnostics.py
```diff
@@ -1555,7 +1562,14 @@ def _bayesian_r2(mu_pred, scale, scale_kind, circular):
...
 
         var_y_est = np.var(mu_pred, axis=1, ddof=1)
-        return var_y_est / (var_y_est + scale)
+        r2 = var_y_est / (var_y_est + scale)
+        if np.any((r2 < 0) | (r2 > 1)):
+            raise ValueError(
+                "Computed R² outside [0, 1]. This usually means the Bernoulli "
+                "pseudo-variance fallback was applied to a non-probability `mu_pred`; "
+                "pass the modelled residual `scale` for continuous models."
+            )
```
jordandeklerk, line 1573:
```suggestion
            raise ValueError(
                "Computed R-squared outside [0, 1]. The Bernoulli pseudo-variance "
                "`mean(mu_pred * (1 - mu_pred))` was used because `scale` was not provided, "
                "but `mu_pred` are not probabilities in [0, 1]. Pass the modelled residual "
                "`scale` for continuous models."
            )
```

jordandeklerk (commented):
Thanks for raising this! Left some pedagogical suggestions. Let me know what you think.

tests/test_metrics.py
```diff
@@ -99,6 +99,24 @@ def test_residual_r2_invalid_shapes():
...
+def test_bayesian_r2_no_scale_warns():
+    # Bernoulli-appropriate mu_pred in [0, 1], scale omitted -> Tjur fallback + warning.
+    rng = np.random.default_rng(0)
+    mu_pred = rng.uniform(0.1, 0.9, size=(200, 40))
+    with pytest.warns(UserWarning, match="Bernoulli"):
+        r2 = array_stats.bayesian_r2(mu_pred)
+    assert np.all((r2 >= 0) & (r2 <= 1))
+
+
+def test_bayesian_r2_continuous_no_scale_errors():
```
jordandeklerk, line 111: This is going to fail because the warning gets caught before the `ValueError`. Since we're explicitly trying to raise the `ValueError` here, we could just decorate the test with `@pytest.mark.filterwarnings("ignore::UserWarning")` here.

CamSalda: You are right. Sure! Change applied.

CamSalda:
> Thanks for raising this! Left some pedagogical suggestions. Let me know what you think.

Thanks for the feedback @jordandeklerk. Lesson learned on the annotations and the format,

jordandeklerk (approved):
LGTM! Thanks again for this.

OriolAbril (commented):
I see I was too late to the party, but I'd like to propose a different solution and potentially a follow up. The docstring was already clear on `scale=None` meaning computing the binary data version, there was also an example in the docstring showing this. Consequently, I don't like showing a warning as it is a valid and desired behaviour. It probably doesn't make sense to have the binary as the default though because people who have not read (or only partially read) the docstring will be able to execute the function and get unexpectedly wrong results.

We have discussed a different placeholder for "no user provided value" in the context of rcParam defaults, and I think here could also be a good place to use it. That is, we define a custom "unset" object (like in [arviz-plots](https://github.com/arviz-devs/arviz-plots/blob/9eb22b042db03617e2a67666c8caf9cd3d9273ba/src/arviz_plots/backend/bokeh/core.py#L37)) and define the function as `scale=unset` with docstring `scale : str or None`. Users can then use `scale=None` for logistic regressions, and `scale="var_name"` otherwise, both paths triggering no warning, but not providing `scale` will error out instead of silently going through the binary data path.

Another option could be to make both `pred_mean` and `scale` required parameters with no default. As far as I can tell, the only valid use of `bayesian_r2(idata)` skipping both these arguments is on an InferenceData from a binary regression and with `p`/`mu` being the only variable in the posterior, which I think is an ok case to break and require explicit passing of `bayesian_r2(idata, "p", scale=None)`

I am leaning more for the 2nd option but would like to have extra input too

CamSalda:
> I see I was too late to the party, but I'd like to propose a different solution and potentially a follow up. The docstring was already clear on `scale=None` meaning computing the binary data version, there was also an example in the docstring showing this. Consequently, I don't like showing a warning as it is a valid and desired behaviour. It probably doesn't make sense to have the binary as the default though because people who have not read (or only partially read) the docstring will be able to execute the function and get unexpectedly wrong results.
> 
> We have discussed a different placeholder for "no user provided value" in the context of rcParam defaults, and I think here could also be a good place to use it. That is, we define a custom "unset" object (like in [arviz-plots](https://github.com/arviz-devs/arviz-plots/blob/9eb22b042db03617e2a67666c8caf9cd3d9273ba/src/arviz_plots/backend/bokeh/core.py#L37)) and define the function as `scale=unset` with docstring `scale : str or None`. Users can then use `scale=None` for logistic regressions, and `scale="var_name"` otherwise, both paths triggering no warning, but not providing `scale` will error out instead of silently going through the binary data path.
> 
> Another option could be to make both `pred_mean` and `scale` required parameters with no default. As far as I can tell, the only valid use of `bayesian_r2(idata)` skipping both these arguments is on an InferenceData from a binary regression and with `p`/`mu` being the only variable in the posterior, which I think is an ok case to break and require explicit passing of `bayesian_r2(idata, "p", scale=None)`
> 
> I am leaning more for the 2nd option but would like to have extra input too

@OriolAbril Here is the [issue](https://github.com/arviz-devs/arviz-stats/issues/390). I will work on the PR to address this comment. I fully agree with your proposed second option.

jordandeklerk:
> I see I was too late to the party, but I'd like to propose a different solution and potentially a follow up. The docstring was already clear on `scale=None` meaning computing the binary data version, there was also an example in the docstring showing this. Consequently, I don't like showing a warning as it is a valid and desired behaviour. It probably doesn't make sense to have the binary as the default though because people who have not read (or only partially read) the docstring will be able to execute the function and get unexpectedly wrong results.
> 
> We have discussed a different placeholder for "no user provided value" in the context of rcParam defaults, and I think here could also be a good place to use it. That is, we define a custom "unset" object (like in [arviz-plots](https://github.com/arviz-devs/arviz-plots/blob/9eb22b042db03617e2a67666c8caf9cd3d9273ba/src/arviz_plots/backend/bokeh/core.py#L37)) and define the function as `scale=unset` with docstring `scale : str or None`. Users can then use `scale=None` for logistic regressions, and `scale="var_name"` otherwise, both paths triggering no warning, but not providing `scale` will error out instead of silently going through the binary data path.
> 
> Another option could be to make both `pred_mean` and `scale` required parameters with no default. As far as I can tell, the only valid use of `bayesian_r2(idata)` skipping both these arguments is on an InferenceData from a binary regression and with `p`/`mu` being the only variable in the posterior, which I think is an ok case to break and require explicit passing of `bayesian_r2(idata, "p", scale=None)`
> 
> I am leaning more for the 2nd option but would like to have extra input too

I think the second option makes sense and is more natural in this case. Appreciate the feedback!

PR #391, opened 2026-07-22
Make pred_mean and scale required in bayesian_r2

### LINKED ISSUE #390: Make `pred_mean` and `scale` required in `bayesian_r2`; drop the Bernoulli fallback warning
#389 Closed https://github.com/arviz-devs/arviz-stats/issues/388

However, follow-up discussions concluded that the real
problem is that `scale` *defaults* to `None`, so a user who has not read the docstring gets the
binary path (and wrong results) for a continuous model without asking for it.

To solve the root problem, make `pred_mean` and `scale` **required parameters with no default** in
`bayesian_r2`. 

`scale=None` stays a valid, explicit value meaning "binary/Bernoulli, compute the
pseudo-variance internally".

CamSalda:
## Description

Follow-up to #389. That PR emitted a `UserWarning` when `scale` was omitted in `bayesian_r2`
(silently taking the Bernoulli pseudo-variance path). Review feedback noted that `scale=None`
meaning "binary/Bernoulli model" is valid and documented, so warning on it is undesirable. The
actual problem is that `scale` *defaults* to `None`, so a user who has not read the docstring
gets the binary path (and wrong results) for a continuous model without asking for it.

This implements the proposed option: make `pred_mean` and `scale` **required** parameters. The
binary path becomes opt-in via an explicit `scale=None`.

Also drops the `scale=None` default on the array-level `array_stats.bayesian_r2` `(src/arviz_stats/base/array.py)` so scale is required there too, keeping the lower-level API consistent with the public function.

Closes https://github.com/arviz-devs/arviz-stats/issues/390

### DIFF
--- a/src/arviz_stats/base/array.py
+++ b/src/arviz_stats/base/array.py
@@ -849,7 +849,7 @@ class BaseArray(_DensityBase, _DiagnosticsBase):
             **kwargs,
         )
 
-    def bayesian_r2(self, mu_pred, scale=None, scale_kind="sd", circular=False):
+    def bayesian_r2(self, mu_pred, scale, scale_kind="sd", circular=False):
         """Compute Bayesian R² for regression models."""
         r2_ufunc = make_ufunc(
             self._bayesian_r2,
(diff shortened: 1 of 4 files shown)


OriolAbril (approved):
Thanks!

PR #392, opened 2026-07-26
fix: add helpful error message when hdi_prob is passed instead of prob

### LINKED ISSUE #384: `az.hdi()` raises TypeError on `hdi_prob` kwarg - renamed to `prob` without a deprecation path
## Description

`az.hdi()` (which now delegates to `arviz_stats.visualization.hdi()`) no longer accepts the `hdi_prob` keyword argument that was the standard, documented way to specify the credible interval probability in ArviZ <1.0. Calling it with `hdi_prob=` now raises a TypeError instead of a deprecation warning:

```
    TypeError: hdi got an unexpected keyword argument: 'hdi_prob'
```

This surfaced as a user-facing bug report against PyMC (pymc-devs/pymc#8365). Since the failure happens well after successful NUTS sampling, in unrelated post-processing code, and does not indicate that the root cause is an ArviZ/arviz-stats API rename rather than a PyMC or user error.

## Root cause

```
`arviz_stats/visualization.py::hdi()` defines the parameter as `prob`:

    def hdi(
        data,
        prob=None,
        dim=None,
        ...
    ):
```

Any caller still using the old `hdi_prob=` name (which is what essentially all existing ArviZ tutorials, StackOverflow answers, and user code use) hits
the `**kwargs` catch-all and gets a generic TypeError with no mention of the rename, no pointer to the new name, and no deprecation period.

## Reproduction

```
    import arviz as az
    import numpy as np

    data = np.random.default_rng().normal(size=2000)
    az.hdi(data, hdi_prob=0.95)  # TypeError: hdi got an unexpected keyword argument: 'hdi_prob'
```

Correct call with the new API:

```

    az.hdi(data, prob=0.95)  # works
```

## Environment

- arviz 1.2.0 / arviz-stats 1.2.0 / arviz-base 1.2.0 (via `conda list` in a
  fresh miniforge env, see pymc-devs/pymc#8365 for full env dump)
- Python 3.12.13, Windows 11

## Suggested fix

Accept `hdi_prob` as a deprecated alias for `prob` for at least one release cycle, emitting a `DeprecationWarning` pointing to the new name, rather than silently landing in `**kwargs` and producing an opaque TypeError. This is a very common breaking change to hit silently since `hdi_prob` appears throughout years of existing ArviZ documentation and tutorials that haven't been updated yet.

sheikhayaan:
## Summary
Fixes #384

## What problem does this fix?
When users pass `hdi_prob=0.95` (the old argument name from ArviZ <1.0),
they get a generic TypeError with no guidance. This is confusing because
`hdi_prob` was the standard documented argument for years.

## What did I change?
Added a specific check for `hdi_prob` in kwargs that raises a helpful
TypeError pointing users to the migration guide and the new argument name.

## Before this fix:
TypeError: hdi got an unexpected keyword argument: 'hdi_prob'

## After this fix:
TypeError: hdi got an unexpected keyword argument: 'hdi_prob'. 
The 'hdi_prob' argument was renamed to 'prob' in arviz-stats 1.0. 
Use 'prob' instead (e.g. prob=0.95). 
See the migration guide: https://python.arviz.org/en/stable/user_guide/migration_guide.html

## How did I test it?
Manually tested with:
import arviz_stats as azs
import numpy as np
data = np.random.default_rng().normal(size=2000)
azs.hdi(data, hdi_prob=0.95)

### DIFF
--- a/src/arviz_stats/base/array.py
+++ b/src/arviz_stats/base/array.py
@@ -128,11 +128,18 @@ class BaseArray(_DensityBase, _DiagnosticsBase):
             func_kwargs.update(kwargs)
         else:
             extra_kwargs = list(kwargs.keys())
+            if "hdi_prob" in kwargs:
+                raise TypeError(
+                    "hdi got an unexpected keyword argument: 'hdi_prob'. "
+                    "The 'hdi_prob' argument was renamed to 'prob' in arviz-stats 1.0. "
+                    "Use 'prob' instead (e.g. prob=0.95). "
+                    "See the migration guide: "
+                    "https://python.arviz.org/en/stable/user_guide/migration_guide.html"
+                )
             if len(extra_kwargs) == 1:
                 raise TypeError(f"hdi got an unexpected keyword argument: '{extra_kwargs[0]}'")
             if len(extra_kwargs) > 1:
                 raise TypeError(f"hdi got unexpected keyword arguments: {extra_kwargs}")
-
         if method == "multimodal_sample":
             func_kwargs["from_sample"] = True
 
--- a/src/arviz_stats/base/array.py
+++ b/src/arviz_stats/base/array.py
@@ -133,8 +133,8 @@ class BaseArray(_DensityBase, _DiagnosticsBase):
                     "hdi got an unexpected keyword argument: 'hdi_prob'. "
                     "The 'hdi_prob' argument was renamed to 'prob' in arviz-stats 1.0. "
                     "Use 'prob' instead (e.g. prob=0.95). "
-                    "See the migration guide: "
-                    "https://python.arviz.org/en/stable/user_guide/migration_guide.html"
+                    "See the HDI Function Changes section in the migration guide: "
+                    "https://python.arviz.org/en/stable/user_guide/migration_guide.html#hdi-function-changes"
                 )
             if len(extra_kwargs) == 1:
                 raise TypeError(f"hdi got an unexpected keyword argument: '{extra_kwargs[0]}'")
--- a/tests/test_minimal.py
+++ b/tests/test_minimal.py
@@ -446,5 +446,5 @@ def test_thin_invalid_factor(data_c0d1):
 
 def test_hdi_invalid_kwarg(data_c0d1):
     """Test that passing hdi_prob raises a helpful error message."""
-    with pytest.raises(TypeError, match="hdi_prob.*renamed.*prob"):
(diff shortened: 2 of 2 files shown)


OriolAbril (commented):
The last two messages on the issue being targetted include

> It might be easier to update the error currently being triggered to add a link to the migration guide.

> I think probably a link to the migration guide would be helpful rather than an explicit explanation in the TypeError considering there are several changes that happened after ArviZ 1.0 that are somewhat related.

Moreover, linking to the migration guide as a whole is not much of a help, even more so because (as also mentioned in the issue) there isn't anything about `hdi` in it.

---

As mentioned in the comment, I think it would be better to add a section about hdi on the migration guide and link directly to that section for any unexpected argument encountered.

sheikhayaan:
"Thanks for the feedback! I'll add an hdi section to the migration guide and update the error message to link directly to that new section. Will push the changes shortly."

sheikhayaan:
"Thanks for the feedback! I see the migration guide lives in the main arviz repo at python.arviz.org/en/stable/user_guide/migration_guide.html. My plan is to:

Add an hdi section to the migration guide in the arviz repo
Update the error message in array.py to link directly to that new section

Does this approach sound right? Should I open a separate PR in the arviz repo for the migration guide update, or keep everything in one PR here?"

sheikhayaan:
"Hi @OriolAbril — just checking in on this. I'm ready to proceed with adding an hdi section to the migration guide and updating the error message to link directly to that section. Should I open a separate PR in the main arviz repo for the migration guide, or include both changes here?"

OriolAbril:
You can't make changes to the migration guide through a PR to arviz-stats, it can only be a PR to the `arviz` repo which is what hosts the migration guide. Here you'll need to update the error to link there once it has been updated

sheikhayaan:
"Thanks for clarifying! I'll open a separate PR in the arviz repo to add the hdi section to the migration guide, then update this PR to link directly to that new section once it's merged."

sheikhayaan:
"Hi @OriolAbril — I've opened a separate PR in the arviz repo to add the hdi section to the migration guide: arviz-devs/arviz#2601. Once that's merged, I'll update the error message in this PR to link directly to that new section. Please let me know if the migration guide content looks good."

sheikhayaan:
"Updated the error message to link directly to the new HDI Function Changes section in the migration guide (arviz-devs/arviz#2601). Ready for re-review when the migration guide PR is merged."

PR #399, opened 2026-08-03
Fix reloo k threshold

### LINKED ISSUE #398: Default value of `k_threshold` in `reloo` is inconsistent with the documentation
When no value is provided by the user, `k_threshold` is set to -inf, which leads to all data points being refitted.

https://github.com/arviz-devs/arviz-stats/blob/73e9158c3c5a3e30bca0736a4cd55ac711395d4c/src/arviz_stats/loo/reloo.py#L23

The documentation, however, says:

> Defaults to $\min(1 - 1/\log_{10}(S), 0.7)$, where S is the number of samples.

I see that there is code that implements this behaviour but it is only executed if `k_threshold` is `None`:

https://github.com/arviz-devs/arviz-stats/blob/73e9158c3c5a3e30bca0736a4cd55ac711395d4c/src/arviz_stats/loo/reloo.py#L192-L193

So setting `k_threshold` to `None` in line 23 should resolve the issue.

sheikhayaan:
Fixes #398

**What changed:**
Changed the default value of `k_threshold` from `-np.inf` to `None` in `reloo.py` to match the documentation, as requested in the issue.

### DIFF
--- a/src/arviz_stats/base/array.py
+++ b/src/arviz_stats/base/array.py
@@ -128,18 +128,11 @@ class BaseArray(_DensityBase, _DiagnosticsBase):
             func_kwargs.update(kwargs)
         else:
             extra_kwargs = list(kwargs.keys())
-            if "hdi_prob" in kwargs:
-                raise TypeError(
-                    "hdi got an unexpected keyword argument: 'hdi_prob'. "
-                    "The 'hdi_prob' argument was renamed to 'prob' in arviz-stats 1.0. "
-                    "Use 'prob' instead (e.g. prob=0.95). "
-                    "See the migration guide: "
-                    "https://python.arviz.org/en/stable/user_guide/migration_guide.html"
-                )
             if len(extra_kwargs) == 1:
                 raise TypeError(f"hdi got an unexpected keyword argument: '{extra_kwargs[0]}'")
             if len(extra_kwargs) > 1:
                 raise TypeError(f"hdi got unexpected keyword arguments: {extra_kwargs}")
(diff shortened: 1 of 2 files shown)


src/arviz_stats/base/array.py
jordandeklerk, line 1: These changes are not relevant to this PR. Remove these.

src/arviz_stats/loo/reloo.py
jordandeklerk: Note that since we're changing the default behavior of `reloo`, it might be a good idea to be mroe explicit here about how to enable automatic refitting for all values.


```suggestion
    As an extreme case, setting ``k_threshold=-np.inf`` will flag every observation and
    make ``reloo`` perform the whole exact LOO-CV, requiring one refit per observation.
    This is not generally recommended nor intended, however, if needed, this function can
    be used to achieve the result.
```

jordandeklerk (changes requested):
Thanks for this! We need to remove the irrelevant changes for the HDI error from this PR.

src/arviz_stats/base/array.py
sheikhayaan: "Thanks for the review! I've removed the unrelated array.py changes and updated the docstring with your suggested text. Please take another look when you get a chance."

src/arviz_stats/loo/reloo.py
sheikhayaan: "Done — updated the docstring with your suggested text."

jordandeklerk (approved):

PR #406, opened 2026-08-15
chore: add livedocs tox environment matching arviz-base

sheikhayaan:
## Summary
Part of DevOps standardization across the ArviZ ecosystem.

## What does this add?
Adds a `livedocs` tox environment to arviz-stats matching the 
one already present in arviz-base. This allows contributors to 
build a live preview of HTML docs with automatic updates during 
development.

## Why?
arviz-base already has this environment but arviz-stats was 
missing it. This is part of ongoing effort to standardize 
development tooling across all ArviZ repositories.

## Related
- arviz-base reference: https://github.com/arviz-devs/arviz-base/blob/main/tox.ini
- GSoC 2025/2026 DevOps standardization project

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -116,7 +116,6 @@ extras =
     numba
 commands =
     sphinx-build -d "{toxworkdir}/docs_doctree" docs/source "{toxworkdir}/docs_out" --color -v -bhtml
-    
 
 [testenv:livedocs]
 description = Build live preview of HTML docs with automatic updates
@@ -126,8 +125,6 @@ extras =
     xarray
 commands =
     sphinx-autobuild -d "{toxworkdir}/docs_doctree" docs/source "{toxworkdir}/docs_out" --color -v -bhtml
-
-
 
 [testenv:cleandocs]
 description = Clean HTML doc build outputs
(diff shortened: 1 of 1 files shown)


tox.ini
OriolAbril: The new task is relevant for contributors so it should be listed here too

tox.ini
```diff
@@ -118,6 +117,15 @@ extras =
 commands =
     sphinx-build -d "{toxworkdir}/docs_doctree" docs/source "{toxworkdir}/docs_out" --color -v -bhtml
 
+[testenv:livedocs]
+description = Build live preview of HTML docs with automatic updates
+base = docs
+extras =
+    doc
+    xarray
```
OriolAbril:
```suggestion
deps = sphinx-autobuild
```

extras are inherited from the `docs` task. In arviz-stats they are updated so the `sphinx-autobuild` dependency is included, but it is done incorrectly. There is no `sphinx-autobuild` extras defined in pyproject, and this being purely a single dev related dependency I think it is better to add it here.

I am quite sure doing this will get the task to work but please try it after doing the change.

OriolAbril (changes requested):

tox.ini
sheikhayaan: "Done — added livedocs to the dev labels and switched to deps = sphinx-autobuild. Let me know if anything else needs fixing."

PR #407, opened 2026-08-15
chore: add --hook-stage manual to pre-commit run matching arviz-base

sheikhayaan:
## Summary
Part of DevOps standardization across the ArviZ ecosystem.

## What does this change?
Adds `--hook-stage manual` flag to pre-commit run command in 
tox check environment, matching the configuration already 
present in arviz-base.

## Why?
arviz-base uses `--hook-stage manual` but arviz-stats was 
missing this flag. This ensures manual-stage hooks are also 
run during CI checks, improving consistency across repos.

## Related
- arviz-base reference: https://github.com/arviz-devs/arviz-base/blob/main/tox.ini
- GSoC DevOps standardization project

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -102,7 +102,7 @@ deps =
 skip_install = true
 commands =
     pre-commit install
-    pre-commit run --all-files --show-diff-on-failure
+    pre-commit run --all-files --show-diff-on-failure --hook-stage manual
     python -m build
 
 [testenv:docs]


OriolAbril:
This is already being added in https://github.com/arviz-devs/arviz-stats/pull/383. Moreover, for full context, the `hook-stage manual` part is because we don't want docstub to run automatically when comitting, only when running `tox -e check`. While not harmful, adding this without docstub being present doesn't add anything either

sheikhayaan:
"Thank you for the context — I wasn't aware of PR #383 or the docstub dependency. I understand now that these settings have specific reasons behind them, not just consistency for its own sake. I'll review existing open issues and PRs more carefully before submitting changes going forward."

Issue #409, opened 2026-08-18
Regression in `uniformity_test`: truncated Cauchy combination deviates from its reference and crashes on valid input


jarotter:
I ran into this issue through a Claude agent with very strict harnesses that decided to write a unit test of whether the test for uniformity test exposed through `BaseDataArray.uniformity_test` was calibrated. 

## Summary
The `pot_c` and `prit_c` methods implement (I believe — no citation appears in the code) the POT-C and PRIT-C tests of Tesso & Vehtari, ["LOO-PIT predictive model checking"](https://arxiv.org/abs/2603.02928). These aggregate pointwise p-values with the *truncated Cauchy combination* (TCCT; Chen, Xu & Gao, [arXiv:2506.12489](https://arxiv.org/abs/2506.12489)), whose statistic — Eq. (24) in Tesso & Vehtari — is the truncated sum divided by the **total** count n:

    T = (1/n) · Σᵢ tan{(0.5 − pᵢ)π} · 𝕀(pᵢ < 0.5)

The original implementation (#316) computed exactly this:

```python
mask = (ps < 0.5).astype(float)
cauchy_mean = np.mean(cauchy_vals * mask)      # truncated sum / n  ✔ Eq. (24)
```

The refactor in #352 ("Clean uniform test implementation") replaced it with:

```python
cauchy_mean = np.mean(cauchy_vals[idx])        # truncated sum / #retained
```

Masking-then-mean and indexing-then-mean are not the same operation. The change has two consequences:

1. **The statistic roughly doubles.** The retained count is ≈ n/2 under the null, so the denominator change inflates the statistic, and with it the rejection rate (measurements below).
2. **An undefined case appears, guarded by a crash.** When every pointwise p-value is ≥ 0.5 — input *more* uniform-looking than typical — the subset is empty and its mean is 0/0. #352 guards this with a `ValueError` (and a unit test, `test_truncated_no_values_below_half_raises`, that enshrines it). Under Eq. (24) the case is perfectly well-defined — an empty sum gives T = 0, the least suspicious possible value, p = 0.5 — and neither reference paper has an error case here. The crash is deterministic on, e.g., an evenly spaced grid:

   ```python
   grid = (np.arange(300) + 0.5) / 300          # perfectly uniform PITs
   ds = xr.Dataset({"pit": (("sample",), grid)})
   ds.azstats.uniformity_test(dim="sample")     # ValueError
   ```

[...]

PR #352, opened 2026-04-10
Clean uniform test implementation

aloctavodia:
Add `_cauchy_combination` function to reuse logic, removes unused eps

PR #410, opened 2026-08-20
Fix truncated Cauchy combination: restore correct denominator (#409)

jarotter:
The refactor in #352 changed the truncated Cauchy combination from a masked mean (truncated sum divided by the total number of terms, as in Eq. 24 of arXiv:2603.02928) to a mean over the retained subset, roughly doubling the statistic and inflating the rejection rate of pot_c and prit_c under the null. It also made the no-retained-terms case (every pointwise p >= 0.5) undefined, guarded by a ValueError that a single too-uniform column could trigger, aborting whole batched calls.

Restore the Eq. 24 semantics via `np.where` (avoiding the inf * 0 hazard of the original mask multiplication), so the empty-truncation case is simply a statistic of 0 (p = 0.5) and the error case disappears. Update the test helper and expectations accordingly, and add regression guards: grid input returns 0.5 instead of raising, one such column no longer aborts a batched call, and the attained size under the null stays below the regressed level.

### DIFF
--- a/tests/test_uniformity.py
+++ b/tests/test_uniformity.py
@@ -290,8 +290,7 @@ class TestUniformityTestGeneral:
 
     @pytest.mark.parametrize("method", ["pot_c", "prit_c"])
     def test_attained_size_under_null(self, method):
-        """Guard against the denominator regression (see #409).
-        """
+        """Guard against the denominator regression (see #409)."""
         rng = np.random.default_rng(42)
         x = rng.uniform(size=(10_000, 100))
         p_values, *_ = array_stats.uniformity_test(x, axis=-1, method=method)
--- a/tests/test_uniformity.py
+++ b/tests/test_uniformity.py
@@ -13,12 +13,7 @@ def _cauchy_combination(ps, truncate=True):
     ps = np.asarray(ps, dtype=float)
     cauchy_vals = np.tan((0.5 - ps) * np.pi)
     if truncate:
-        idx = ps < 0.5
-        if not np.any(idx):
-            raise ValueError(
-                "Cannot compute truncated Cauchy combination test. No p-values below 0.5 found."
-            )
-        cauchy_vals = cauchy_vals[idx]
+        cauchy_vals = np.where(ps < 0.5, cauchy_vals, 0.0)
     mean_val = np.mean(cauchy_vals)
     return 0.5 - np.arctan(mean_val) / np.pi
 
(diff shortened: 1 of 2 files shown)


tests/test_uniformity.py
```diff
@@ -274,6 +271,45 @@ def test_3d_batched(self, method):
...
+    def test_attained_size_under_null(self, method):
+        """Guard against the denominator regression (see #409).
+
+        On i.i.d. uniform input the rejection rate at alpha = 0.05 is
+        ~0.05-0.07 with the Eq. 24 denominator, but ~0.08-0.10 when the
+        truncated sum is divided by the retained count instead of n. The
+        bounds deliberately allow the method-inherent inflation above
+        0.05 (see the endnote of #409) and sit ~5 Monte-Carlo standard
+        errors from both the correct and the regressed values, so the
+        test stays safe even if the numpy Generator bit-stream changes.
```
aloctavodia:
```suggestion
        """Guard against the denominator regression (see #409).
```

tests/test_uniformity.py
```diff
@@ -274,6 +271,45 @@ def test_3d_batched(self, method):
         assert shapley.shape == (2, 3, 80)
         assert shapley_unsorted.shape == (2, 3, 80)
 
+    @pytest.mark.parametrize("method", ["pot_c", "prit_c"])
+    def test_all_pointwise_ps_above_half(self, method):
+        """An evenly spaced grid makes every pointwise p-value >= 0.5.
+
+        The truncated statistic is then an empty sum, i.e. 0, so the
+        combined p-value is exactly 0.5 (no evidence against uniformity)
+        and the call must not raise.
```
aloctavodia:
```suggestion
```

aloctavodia (approved):

tests/test_uniformity.py
```diff
@@ -274,6 +271,33 @@ def test_3d_batched(self, method):
...
+        rng = np.random.default_rng(0)
+        x = np.vstack([rng.uniform(size=(3, 300)), (np.arange(300) + 0.5) / 300])
+        p_value, *_ = array_stats.uniformity_test(x, axis=-1, method=method)
+        assert p_value.shape == (4,)
+        assert_allclose(p_value[-1], 0.5, atol=1e-10)
+
+    @pytest.mark.parametrize("method", ["pot_c", "prit_c"])
+    def test_attained_size_under_null(self, method):
+        """Guard against the denominator regression (see #409).
+        """
```
aloctavodia:
```suggestion
        """Guard against the denominator regression (see #409)."""
```

tests/test_uniformity.py
```diff
@@ -274,6 +271,32 @@ def test_3d_batched(self, method):
         assert shapley.shape == (2, 3, 80)
         assert shapley_unsorted.shape == (2, 3, 80)
 
+    @pytest.mark.parametrize("method", ["pot_c", "prit_c"])
+    def test_all_pointwise_ps_above_half(self, method):
+        """An evenly spaced grid makes every pointwise p-value >= 0.5.
+        """
```
aloctavodia:
```suggestion
        """An evenly spaced grid makes every pointwise p-value >= 0.5."""
```

aloctavodia:
Thanks a lot @jarotter!

PR #411, opened 2026-08-20
Add PyMC and Bambi support for `loo_subsample`

jordandeklerk:
Adds a `model` arg to `loo_subsample` and `update_subsample` that auto-builds the pointwise log-likelihood function from a PyMC or Bambi model via a new `ll_from_pymc` helper, so `method="plpd"` no longer requires hand-writing a `log_lik_fn` (for `method="lpd"` the model is optional and recomputes the approximation instead of using the stored log-likelihood values). 

Also added `pymc_helper.py` which has both the moment matching and subsampling shared PyMC helpers. We will probably be adding more automatic PyMC extraction for other methods in the future so I figured we should just have a central place for this.

### DIFF
--- a/src/arviz_stats/loo/loo_subsample.py
+++ b/src/arviz_stats/loo/loo_subsample.py
@@ -37,6 +37,7 @@ def loo_subsample(
     param_names=None,
     log=True,
     log_jacobian=None,
+    model=None,
 ):
     r"""Compute PSIS-LOO-CV using sub-sampling.
 
@@ -90,7 +91,8 @@ def loo_subsample(
         Method used for approximating the pointwise log predictive density:
 
         - ``lpd``: Use standard log predictive density approximation (default)
-        - ``plpd``: Use point log predictive density approximation which requires a ``log_lik_fn``.
+        - ``plpd``: Use point log predictive density approximation which requires a ``log_lik_fn``
+          or a ``model`` from which one is built.
     thin: int or str, optional
         Thinning factor for posterior draws. Can be an integer to thin by that factor,
         "auto" to automatically determine thinning based on bulk and tail ESS, or None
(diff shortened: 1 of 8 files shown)


src/arviz_stats/loo/loo_subsample.py
```diff
@@ -432,6 +461,11 @@ def update_subsample(
     log: bool, optional
         Whether the ``log_lik_fn`` returns log-likelihood (True) or likelihood (False).
         Default is True.
+    model : Model, optional
+        A model object. Currently supported models are PyMC and Bambi.
+        If provided and ``log_lik_fn`` is not, it will be used to auto-build ``log_lik_fn``,
+        so it does not have to be constructed manually. ``param_names`` is ignored in
+        this case.
```
aloctavodia: Should we have a warning if both have passed?

src/arviz_stats/loo/loo_subsample.py
```diff
@@ -498,18 +532,35 @@ def update_subsample(
...
+        re_center_intercept = getattr(model, "_re_center_intercept", None)
+        if re_center_intercept is not None:
+            data = re_center_intercept(data)
+            model = model.backend.model
+
+        log_lik_fn = ll_from_pymc(data, model=model, var_name=loo_inputs.var_name)
+        param_names = None
+
+    if method == "plpd" and log_lik_fn is None:
+        raise ValueError("log_lik_fn or model must be provided when method='plpd'")
```
aloctavodia: This code is duplicated

src/arviz_stats/loo/loo_subsample.py
```diff
@@ -498,18 +532,35 @@ def update_subsample(
...
     if reff is None:
         reff = _get_r_eff(data, loo_inputs.n_samples)
 
+    # Auto-build the log likelihood function from the model if provided
+    if model is not None and log_lik_fn is None:
+        if not log:
+            raise ValueError("log must be True when log_lik_fn is auto-built from model.")
+        from arviz_stats.loo.loo_subsample_helper import ll_from_pymc
+
+        ## if model is Bambi's, re-center the intercept so it matches the PyMC value variable
```
aloctavodia:
```suggestion
        # if model is Bambi's, re-center the intercept so it matches the PyMC value variable
```

src/arviz_stats/loo/loo_subsample.py
jordandeklerk: Good point, fixed. Moved this into the helper file to avoid any extra duplication.

aloctavodia (approved):

PR #420, opened 2026-08-31
Monte Carlo calibrate uniformity_test p-values (#419)

### LINKED ISSUE #419: uniformity_test (pot_c) rejects about twice the nominal rate at alpha = 0.05
The default `pot_c` method of `uniformity_test` seems anti-conservative. On exactly uniform data it rejects at alpha = 0.05 about 9-11% of the time.

Reproduction (arviz-stats 1.2.0, also current `main`):

```python
import numpy as np
from arviz_stats.base import array_stats

rng = np.random.default_rng(1)
for n in (200, 500, 1000):
    rej = sum(array_stats.uniformity_test(rng.random(n))[0] < 0.05 for _ in range(1000))
    print(n, rej / 1000)
# 200  0.088
# 500  0.107
# 1000 0.106
```

Same result with randomized PIT values built from discrete SBC ranks, so it is not an input issue.

The existing tests only check a single sample (`p > 0.05` for one uniform draw), so this does not show up in CI. Maybe the Cauchy combination over strongly dependent order statistics needs a correction, or the docs could note that the p-value is not calibrated at conventional levels.

## Suggested fix

The likely cause: the Cauchy combination is only guaranteed in the far tail, and order statistics are strongly dependent, so at conventional levels like 0.05 the combined p-value comes out too small.

The null here is fully known (order statistics of n uniforms), so the combined statistic can be calibrated exactly by simulation:

1. Simulate K sets of n uniform draws, compute the Cauchy combination statistic for each (one vectorized `betainc` call over a (K, n) array, fast).
2. Report the p-value as the rank of the observed statistic in that null sample.
3. Cache per n, the same way the simultaneous ECDF band already calibrates gamma by simulation.

This makes the reported p-value exact by construction at any level, at the cost of one cached simulation per sample size. A cheaper alternative is a docs note that the p-value is only calibrated in the far tail and conventional thresholds like 0.05 should not be used with it.

Also worth adding a CI test that measures the rejection rate over ~1000 null replications instead of checking a single draw.

CamSalda:
Closes #419.

The Cauchy combination is only guaranteed in the far tail, and the pointwise terms it combines here are strongly dependent. As a result the analytic p-value rejects about twice the nominal rate at alpha = 0.05 (measurements in #419).

Since the null is fully known, `pot_c`, `prit_c` and `piet_c` now re-rank their analytic p-value against 1000 simulated null p-values, computed once and cached per test and sample size.

* Attained size is now close to nominal. A new test checks it for all three methods.
* Extreme p-values keep the analytic value, so far tail gradation is preserved.
* `n_simulations=0` gives the previous behavior. Shapley contributions are unchanged.
* `_mtc_c` is left for a follow-up.

Draft to discuss the default behavior and the number of simulations.

### DIFF
--- a/src/arviz_stats/base/array.py
+++ b/src/arviz_stats/base/array.py
@@ -778,7 +778,11 @@ class BaseArray(_DensityBase, _DiagnosticsBase):
             Method to use for the uniformity test.
             Valid options are pot_c (default), prit_c and piet_c.
         **kwargs
-            Additional keyword arguments.
+            Additional keyword arguments passed to the test function.
+            ``n_simulations`` (default 1000) sets how many null replications
+            are used to Monte Carlo calibrate the p-value (simulated once and
+            cached per sample size); pass 0 for the uncalibrated analytic
+            p-value, which is anti-conservative at conventional levels.
 
         Returns
         -------
--- a/src/arviz_stats/base/density.py
+++ b/src/arviz_stats/base/density.py
@@ -14,6 +14,8 @@ from scipy.stats import binom, hypergeom
 
 from arviz_stats.base.core import _CoreBase
 
+_UNIFORMITY_NULL_CACHE = {}
+
 
 class _DensityBase(_CoreBase):
     """Class with numpy+scipy only density related functions."""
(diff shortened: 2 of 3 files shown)


CamSalda:
Closed as this was already addressed by https://github.com/arviz-devs/arviz-stats/pull/410

PR #424, opened 2026-09-01
Add weight_predictions to docs, add example

aloctavodia:


src/arviz_stats/manipulation.py
```diff
@@ -156,6 +156,21 @@ def weight_predictions(
...
+    --------
+    .. ipython::
+
+        In [1]: import arviz_base as azb
+           ...: import arviz_stats as azs
+           ...: m0 = azb.load_arviz_data("crabs_poisson")
+           ...: m1 = azb.load_arviz_data("crabs_hurdle_nb")
+           ...: model_dict = {"m0": m0, "m1": m1}
+           ...: comp = azs.compare({"m0": m0, "m1": m1})
+           ...: azs.weight_predictions(
```
jordandeklerk, line 170: Should we use a random seed here so docs are reproducible?

jordandeklerk (approved):

src/arviz_stats/manipulation.py
aloctavodia: Not sure, the example is a little bit dumb actually. I added mostly to have something to show.

jordandeklerk: Yeah, fair point. I'm not sure we need to overwork or replace the example IMO. It is still useful for showing how weights from `compare` feed into `weight_predictions`. The seed suggestion was only to keep the rendered output reproducible, so I’m happy to treat that as non-blocking tbh.

PR #437, opened 2026-09-04
Fix jacobian term for p_loo

aloctavodia:
A similar operation is done in _mixture (which I think is the correct thing to do)


jordandeklerk (approved):
Nice catch! LGTM

PR #448, opened 2026-09-11
Only build the summary's LaTeX repr when asked for

### LINKED ISSUE #447: Displaying `az.summary()` in IPython needs jinja2, via `SummaryDataFrame._repr_latex_`
### Describe the bug

With `pip install arviz ipython` (no jinja2), displaying a summary in a notebook shows the table followed by an `ImportError` traceback. `_repr_latex_` always calls `self.to_latex()`, which needs jinja2 through pandas' Styler, and IPython builds every repr on display:

https://github.com/arviz-devs/arviz-stats/blob/a04bc81e2eb4cf3bde3fb68f66ce91bb1d363f81/src/arviz_stats/summary.py#L417-L420

pandas' own `DataFrame._repr_latex_` returns `None` unless `styler.render.repr == "latex"`, so gating the override the same way should resolve it:

```python
if self._fmt_map is None or pd.get_option("styler.render.repr") != "latex":
    return super()._repr_latex_()
```

(Patched locally: no traceback, and the HTML table is unchanged.)

### Steps to reproduce

```python
import numpy as np, arviz as az
from IPython.core.interactiveshell import InteractiveShell

idata = az.from_dict({"posterior": {"mu": np.random.default_rng(0).normal(size=(4, 100))}})
InteractiveShell.instance().display_formatter.format(az.summary(idata))
# ImportError: `Import Jinja2` failed. DataFrame.style requires jinja2. ...
```

### Additional context

arviz 1.3.0, arviz-stats 1.3.2, pandas 3.0.5, IPython 9.17.1, Python 3.12. We hit it first in JupyterLite (Pyodide), where the kernel has only what the notebook installs. The override seems to come from #379.

habakan:
Closes #447.

`_repr_latex_` now returns LaTeX only when `styler.render.repr == "latex"`, as pandas does; the test checks both cases.

Checked: `tests/test_summary.py` and pre-commit pass; with no jinja2, displaying a summary gives no traceback.

### DIFF
--- a/src/arviz_stats/summary.py
+++ b/src/arviz_stats/summary.py
@@ -415,7 +415,6 @@ class SummaryDataFrame(pd.DataFrame):
         return self._display_df().to_html()
 
     def _repr_latex_(self):  # pylint: disable=overridden-final-method
-        # Like pandas, only when asked for: IPython calls every repr, and to_latex needs jinja2.
         if self._fmt_map is None or pd.get_option("styler.render.repr") != "latex":
             return super()._repr_latex_()
         return self.to_latex(escape=True)
--- a/src/arviz_stats/summary.py
+++ b/src/arviz_stats/summary.py
@@ -415,7 +415,8 @@ class SummaryDataFrame(pd.DataFrame):
         return self._display_df().to_html()
 
     def _repr_latex_(self):  # pylint: disable=overridden-final-method
-        if self._fmt_map is None:
+        # Like pandas, only when asked for: IPython calls every repr, and to_latex needs jinja2.
+        if self._fmt_map is None or pd.get_option("styler.render.repr") != "latex":
             return super()._repr_latex_()
         return self.to_latex(escape=True)
 
--- a/tests/test_summary.py
+++ b/tests/test_summary.py
@@ -596,8 +596,6 @@ def test_summary_data_frame():
     assert "0.123" in html
     assert "1.06" in html
 
-    # As for a plain DataFrame, the LaTeX repr is only built when asked for,
-    # since IPython calls every repr on display and to_latex needs jinja2.
     assert sdf._repr_latex_() is None
     with pd.option_context("styler.render.repr", "latex"):
         latex = sdf._repr_latex_()
--- a/tests/test_summary.py
+++ b/tests/test_summary.py
@@ -596,7 +596,11 @@ def test_summary_data_frame():
     assert "0.123" in html
     assert "1.06" in html
 
-    latex = sdf._repr_latex_()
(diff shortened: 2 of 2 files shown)


aloctavodia:
Thanks @habakan. 

@jordandeklerk @OriolAbril do we want this behaviour?

src/arviz_stats/summary.py
```diff
@@ -415,7 +415,7 @@ def _repr_html_(self):  # pylint: disable=overridden-final-method
         return self._display_df().to_html()
 
     def _repr_latex_(self):  # pylint: disable=overridden-final-method
-        if self._fmt_map is None:
+        if self._fmt_map is None or pd.get_option("styler.render.repr") != "latex":
             return super()._repr_latex_()
         return self.to_latex(escape=True)
```
jordandeklerk: I'm pretty sure we can drop this override entirely. Pandas own `_repr_latex_` is already return `self.to_latex()` if `get_option("styler.render.repr") == "latex"` else `None`, and since `self.to_latex()` links to our override below, it picks up the `fmt_map` and `escape=True` on its own, I believe.

tests/test_summary.py
```diff
@@ -596,7 +596,9 @@ def test_summary_data_frame():
     assert "0.123" in html
     assert "1.06" in html
 
-    latex = sdf._repr_latex_()
+    assert sdf._repr_latex_() is None
+    with pd.option_context("styler.render.repr", "latex"):
+        latex = sdf._repr_latex_()
```
jordandeklerk: Could we pin the option for the is `None` check too? `option_context("styler.render.repr", "html")` keeps it from depending on the global state. Then the latex block can just be `assert sdf._repr_latex_() == sdf.to_latex()` rather than repeating the substring checks.

jordandeklerk (changes requested):
Thanks for this @habakan!

@aloctavodia @OriolAbril I got this exact error the other day, so I do think it's relevant. Let me know if I'm missing something.

habakan:
Good catch, thanks @jordandeklerk — you're right that pandas' own `_repr_latex_` already gates on `styler.render.repr` and dispatches through `self.to_latex()`, which resolves to our override. Dropped the override entirely in b8c7c30 and tightened the tests per your suggestion (pinned the `None` check under `option_context`, and asserted `_repr_latex_() == to_latex()` instead of repeating substring checks).

jordandeklerk (approved):
Thanks! I think this looks good, but curious to hear from others before merging.

aloctavodia (approved):
Now I wonder why I added this. I remember something was not working with the LaTeX repr. But maybe it was something different, haha. LGTM!

PR #449, opened 2026-09-14
Add ELPD subclasses 

### LINKED ISSUE #444: Refactor `ELPDData` into type-specific subclasses
## Motivation

`ELPDData` is a single class serving all four ELPD-producing functions:

- **PSIS-LOO-CV** (`loo()`, `loo_approximate_posterior()`)
- **LOO subsampling** (`loo_subsample()`, `update_subsample()`, `loo_moment_match()`)
- **k-fold cross-validation** (`loo_kfold()`)
- **PSIS LFO-CV** (`lfo_cv()` — added in PR #305)

PR #305 brings the total to **29 attributes**, of which only 11 are shared across all sources. The rest are source-specific:

 | Source | Attributes |
 |---|---|
 | **Shared (all 4 sources)** | `kind`, `elpd`, `se`, `p`, `n_samples`, `n_data_points`, `scale`, `warning`, `good_k`, `elpd_i`, `pareto_k` |
 | **LOO** | `log_weights`, `approx_posterior` |
 | **LOO subsample** | `subsample_size`, `subsample_se`, `log_p`, `log_q`, `thin_factor`, `loo_subsample_observations`, `elpd_loo_approx`, `log_jacobian`, `influence_pareto_k`, `n_eff_i` |
 | **LOO kfold** | `n_folds`, `fold_fits`, `p_kfold_i` |
 | **LFO-CV** | `forecast_horizon`, `min_observations`, `refits`, `n_refits`, `p_lfo_i` |

## Problems

1. **Autocomplete noise** — `loo()` users see `subsample_size`, `n_folds`, `forecast_horizon`, etc.
2. **Complex printing** — `__str__` already has `if kind == "loo_kfold"` and `if subsample_size` branches; LFO-CV needs its own.
3. **Growing class** — already flagged `too-many-instance-attributes`; more attributes make this warning unavoidable.

## Proposal

Split `ELPDData` into foru subclasses: `ELPDDataLOO`, `ELPDDataLOOSubsample`, `ELPDDataLOOKFold`, and `ELPDDataLFO`.

Each subclass overrides `__str__` for its own format, removing the `if kind == ...` branching.

## Open questions

- Should `ELPDData` remain a public import, or should users use subclasses exclusively?
- How does `compare()` handle all four types?
- `kind` discriminator vs. `isinstance` checks?
- Migration path for dynamic attribute access?

## Related

 - PR #305 (Add PSIS LFO-CV)
 - Issue #177nd

jordandeklerk:
`ELPDData` had grown to 29 fields with only 11 shared across estimators, and `__str__` branched on `kind`, `subsample_size`, and `n_folds`. This splits it into a keyword-only hierarchy so each function returns an object with just its own attributes:

- `ELPDData` keeps the 11 shared fields.
- `ELPDDataLOO` (`loo`, `loo_i`, `loo_approximate_posterior`, `loo_moment_match`, `reloo`) adds the PSIS fields and declares `p_loo_i`, `influence_pareto_k`, `n_eff_i` instead of assigning them ad hoc.
- `ELPDDataLOOSubsample` subclasses `ELPDDataLOO` and adds the subsampling fields.
- `ELPDDataLOOKFold` and `ELPDDataLFO` add theirs.

Each subclass prints its own header, so the `kind`-based branching in `__str__` is gone. `compare`, `reloo`, `loo_moment_match`, and `update_subsample` use `isinstance` instead of `getattr`/`hasattr`, and `compare` builds its internal frame from the four columns it reads rather than deep-copying whole result objects.

---

Resolves https://github.com/arviz-devs/arviz-stats/issues/444


src/arviz_stats/loo/compare.py
```diff
@@ -252,9 +250,16 @@ def compare(
...
-    ics = pd.DataFrame.from_dict(ics_dict, orient="index")
+    ics = pd.DataFrame(
+        {
+            "elpd": [elpd_data.elpd for elpd_data in ics_dict.values()],
+            "se": [elpd_data.se for elpd_data in ics_dict.values()],
+            "p": [elpd_data.p for elpd_data in ics_dict.values()],
+            "elpd_i": [elpd_data.elpd_i.values.flatten() for elpd_data in ics_dict.values()],
+        },
+        index=names,
+    )
```
jordandeklerk, line 261: For some context here: 

`compare` used to build its working frame with `pd.DataFrame.from_dict(ics_dict, orient="index")`. As far as I can tell, Pandas handles a dict of dataclass instances by calling `dataclasses.asdict` on each one, which expands every field into a column and deep-copies the values (`log_weights`, `pareto_k`, and so on), even though only `elpd`, `se`, `p`, and `elpd_i` are ever read from the frame. 

That was sort of harmless before, but with `fold_fits` now a declared field it would copy every fitted model into a throwaway `DataFrame`, which could potentially be time consuming. Building the four columns explicitly avoids this, I believe.

docs/source/api/index.md
```diff
@@ -91,6 +91,19 @@ you should jump to {ref}`array_stats_api` and read forward.
...
+```{eval-rst}
+.. autosummary::
+   :toctree: generated/
+
+   arviz_stats.ELPDData
+   arviz_stats.ELPDDataLOO
+   arviz_stats.ELPDDataLOOSubsample
+   arviz_stats.ELPDDataLOOKFold
+   arviz_stats.ELPDDataLFO
+```
```
jordandeklerk, line 105: Don't know if we want this? We didn't document it before, but now might seem confusing if you look at the return types of the different LOO functions now that there are subclasses.

aloctavodia: This is good

aloctavodia:
What is the role of `kind` now? Should we keep it? Now it is possible to set kind to something that will not match the class, right? Should we remove it? Or make it a property? 
Everything else looks fine.

jordandeklerk:
> What is the role of `kind` now? Should we keep it? Now it is possible to set kind to something that will not match the class, right? Should we remove it? Or make it a property? Everything else looks fine.

Yeah good point. You could in theory set kind to anything because it was exposed. Made it a property because `compare` still needs to be able to indentify different LOO estimators. So if we removed it, we'd have to just invent something very similar to it anyway. 

This change is going to affect some tests in `arviz-plots` as well. Here https://github.com/arviz-devs/arviz-plots/blob/main/tests/conftest.py#L209 and here https://github.com/arviz-devs/arviz-plots/blob/main/tests/test_plots.py#L487. I'll open a quick change for this soon.

aloctavodia (approved):

PR #453, opened 2026-09-17
Add default `ci_kind` handling in `ci_in_rope`

jordandeklerk:
`ci_in_rope` says that `ci_kind` defaults to `rcParams["stats.ci_kind"]` but never actually enforced that, so leaving it unset always computed the HDI even though the rcParam defaults to `"eti"`. Now resolves `None` from the rcParam like `summary` and the r2 functions do.


src/arviz_stats/summary.py
```diff
@@ -483,7 +483,6 @@ def ci_in_rope(
         Probability for the credible interval. Defaults to ``rcParams["stats.ci_prob"]``.
     ci_kind : {"hdi", "eti"}, optional
         Type of credible interval. Defaults to ``rcParams["stats.ci_kind"]``.
-        If `kind` is stats_median or all_median, `ci_kind` is forced to "eti".
```
aloctavodia, line 486: Is this not true?

jordandeklerk: I thought this was a copy over from the `summary()` docstring where you could pass an explicit `kind` argument tbh. I don't see where you could specify this in `ci_in_rope()` though. Am I missing something? Could this flow through to `ci_in_rope()` somehow?

aloctavodia: Sorry, my bad (I should not review using a cellphone). I read this as being from summary, haha.

aloctavodia (approved):

src/arviz_stats/summary.py
jordandeklerk: No worries!

